PR #19, opened 2021-11-04
Matveibos hometask 1

Matveibos:
Домашняя работа 1

### DIFF
--- a/Matveibos/hometask-1/hello_world.rb
+++ b/Matveibos/hometask-1/hello_world.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+# !/usr/bin/env ruby
+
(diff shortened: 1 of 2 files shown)


ramantechart (changes requested):
Fix your commit history.
Style fixes should be wrapped in one commit.

Matveibos/hometask-1/hello_world.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+# !/usr/bin/env ruby
```
ramantechart: This won't make the file executable.
Try to run `$ ./hello_world.rb` in the hometask folder.

ramantechart (changes requested):

PR #25, opened 2021-11-05
homework_1

Romuashi:
Первая домашка


bedyan (changes requested):
This file is not executable, you need to make it able to execute like this `./hello_world.rb`

bedyan (approved):
GJ :+1: ! Next time try to update existing file instead of deleting and creating new one.

PR #29, opened 2021-11-05
silitski_homework_1

qqqwww123cdv:


jekapulpit (changes requested):
you need to make this file executable
to make sure you've done it right, try to execute it using `./hello_world.rb`

PR #30, opened 2021-11-06
PandaGrom hometask 2

PandaGrom:
Complete hometask 2

### DIFF
--- a/PandaGrom/hometask-2/get_middle_character.rb
+++ b/PandaGrom/hometask-2/get_middle_character.rb
@@ -8,7 +8,7 @@ end
 
 def run_cli
   loop do
-    print 'Enter string: '
+    p 'Enter string: '
     str = gets.chomp
 
     break if str == 'exit!'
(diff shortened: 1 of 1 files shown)


PandaGrom/hometask-2/get_middle_character.rb
```diff
@@ -0,0 +1,24 @@
...
+
+def run_cli
+  loop do
+    print 'Enter string: '
+    str = gets.chomp
+
+    break if str == 'exit!'
+
+    if str.empty?
+      p 'String can not be blank!'
```
nieisnuje, line 17: You used `print` in line 11 and now you are using `p`
Use the same method here to make your code consistent

PandaGrom/hometask-2/get_middle_character.rb
```diff
@@ -0,0 +1,24 @@
...
+    else
+      p "Middle character: #{get_middle(str)}"
```
nieisnuje, line 19: Same as above

nieisnuje (approved):

PR #32, opened 2021-11-06
silitski/hometask 1

qqqwww123cdv:
I made a small program.


jekapulpit (changes requested):
why did you close your previous PR? 😄 
btw it's still not executable

PR #33, opened 2021-11-06
Silitski hometask 2

qqqwww123cdv:
Another small program.


hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+def get_middle(str)
+  if (str.length % 2).zero?
```
jekapulpit: in ruby, we can use `even?` for this

hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,21 @@
...
+    puts(str.slice((str.length / 2 - 1)..(str.length / 2)))
```
jekapulpit: `str.length / 2` appears in your code several times, so how about extracting it into a local variable?

hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+def get_middle(str)
```
jekapulpit: actually, your method is not `getting` the middle, but prints it, so how about making this method actually `return` the middle, and at the end of your loop just printing the result?

jekapulpit (changes requested):
let's try to have decent and accurate commit names, especially it makes sense to think about it when you duplicate your commits 😄 
it's not a blocker, just a piece of advice for the future.

PR #37, opened 2021-11-07
Silitski/hometask2

qqqwww123cdv:
Сompleted second homework.


hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+def get_middle(str)
+  hf = str.length / 2
```
jekapulpit: let's try to name this variable more accurately? `hf` gives not many information 😄

hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,23 @@
...
+  if str.length.even?
```
jekapulpit: we can use ternary operator there and your 5 lines will become 1

hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,23 @@
...
+    str[hf - 1] + str[hf]
```
jekapulpit: and if you'll choose the ternary operator, the range looks better here `str[(hf - 1)..hf]`

jekapulpit (changes requested):

ramantechart (changes requested):
https://github.com/iTechArt-RubyLab/Ruby-Lab#%D0%B7%D0%B0%D0%B4%D0%B0%D0%BD%D0%B8%D0%B5-2

PR #39, opened 2021-11-07
TehanovAnton/hometask-2

TehanovAnton:
`get_middle method` commit: [codewars](https://www.codewars.com/kata/56747fd5cb988479af000028/train/ruby) task
`run_cli method` commit: ruby-lab task

### DIFF
--- a/TehanovAnton/hometask-2/get_middle_charachter.rb
+++ b/TehanovAnton/hometask-2/get_middle_charachter.rb
@@ -1,6 +1,32 @@
+#!/usr/bin/env ruby
 # frozen_string_literal: true
+
+ERROR_MESSAGE = "String can't be blank"
+ENTER_PROMPT = 'Enter string:'
+EXIT = 'exit!'
 
 def get_middle(string)
   half = string.size / 2
   string.length.odd? ? string[half] : string[(half - 1)..half]
 end
+
+def error_message
+  ERROR_MESSAGE
+end
+
+def result(input)
+  "Middle character: #{get_middle(input)}"
+end
+
+def run_cli
+  loop do
+    puts ENTER_PROMPT
+    input = gets.chomp
+
+    exit if input == EXIT
+
+    puts input.empty? ? error_message : result(input)
+  end
+end
+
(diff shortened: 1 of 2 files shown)


TehanovAnton/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+ERROR_MESSAGE = "String can't be blank"
+ENTER_PROMPT = 'Enter string:'
+EXIT = 'exit!'
```
dmokreckiy: @TehanovAnton can you please rationalize usage of constants instead of just plain strings?

TehanovAnton/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,32 @@
...
+
+def error_message
+  ERROR_MESSAGE
+end
+
+def result(input)
+  "Middle character: #{get_middle(input)}"
+end
+
+def run_cli
```
dmokreckiy: @TehanovAnton you already should have a concept of a loop through CodeAcademy lessons. Can you try to rewrite it with that context in mind?

dmokreckiy (dismissed):
LGTM
But check my comments before merging

TehanovAnton/hometask-2/get_middle_charachter.rb
TehanovAnton: @dmokreckiy What do you mean? Do you mean we should not use constants in this case?

TehanovAnton/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,32 @@
...
+  loop do
+    puts ENTER_PROMPT
+    input = gets.chomp
+
+    exit if input == EXIT
```
TehanovAnton: replace exit with break

dmokreckiy (approved):
Really good work!
LGTM

PR #40, opened 2021-11-07
AndrewTehan-hometask 2

AndrewTehan:
[hometask-2](https://github.com/iTechArt-RubyLab/Ruby-Lab/blob/rubylab-2021-2022/README.md#%D0%B7%D0%B0%D0%B4%D0%B0%D0%BD%D0%B8%D0%B5-2)

### DIFF
--- a/AndrewTehan/hometask-2/get_middle_charachter.rb
+++ b/AndrewTehan/hometask-2/get_middle_charachter.rb
@@ -1,13 +1,29 @@
-def return_by_index(string, index)
-  string[index]
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+ENTER_MESSAGE = 'Enter string: '
+ERROR_MESSAGE = "String can't be blank"
+RESULT_MESSAGE = 'Middle character: '
+EXIT_MESSAGE = 'exit!'
+
+def run_cli
+  loop do
+    print ENTER_MESSAGE
+    string = gets.chomp
+
+    exit if string == EXIT_MESSAGE
+
+    if string.empty?
+      puts ERROR_MESSAGE
+    else
+      puts RESULT_MESSAGE + get_middle(string)
+    end
+  end
 end
 
-def return_by_range(string, start_index, end_index)
-  string[start_index..end_index]
+def get_middle(string)
+  middle_index = (string.size - 1) / 2.0
+  string[middle_index.floor..middle_index.round]
 end
 
-def get_middle(string) 
-  size = string.size
-  middle_index = size / 2
-  size.odd? ? return_by_index(string, middle_index) : return_by_range(string, middle_index - 1, middle_index)
-end
+run_cli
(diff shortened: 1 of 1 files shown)


AndrewTehan/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+ENTER_STRING = 'Enter string: '
+ERROR_MESSAGE = "String can't be blank"
+EMPTY_STRING = ''
+EXIT = 'exit!'
```
hannazspot: I'd recommend being more consistent in naming, cause I don't get a point why some of the messages are called as `*_STRING`, others - `*_MESSAGE` and one of them is w/o any prefix in general

AndrewTehan/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,36 @@
...
+  print ENTER_STRING
+  string = gets.chomp
+
+  case string
+  when EXIT then exit
+  when EMPTY_STRING then puts ERROR_MESSAGE
+  else puts get_middle(string)
+  end
+
+  run_cli
```
hannazspot: I'd avoid recursion, it could provide a stack overflow error. I can give a simple example of that situation:
if I run a program once and call this method a lot of times (a huge amount), my laptop will likely run out of stack memory.

You can look through [the article](http://vaidehijoshi.github.io/blog/2014/12/14/to-understand-recursion-you-must-first-understand-recursion/) and get down to this topic deeper.

My point here is to avoid recursion AMAP and use loops instead of it.

AndrewTehan/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,36 @@
...
+end
+
+def return_by_index(string, index)
```
hannazspot: `return_` in methods' names seems weird a bit. One of the main purposes of methods is to do and **return** smth. So I'd say that it's not necessary to put such meanless words in naming from this point of view.

AndrewTehan/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,36 @@
...
+  string[index]
+end
+
+def return_by_range(string, start_index, end_index)
+  string[start_index..end_index]
+end
+
+def get_middle(string)
```
hannazspot, line 24: Could you look at [ruby style guide](https://github.com/rubocop/ruby-style-guide#accessormutator-method-names)? it's highly recommended to avoid using such prefixes as `get_*`, `set_*`, so let's try to follow it

AndrewTehan/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,36 @@
...
+  size = string.size
+  middle_index = size / 2
+  size.odd? ? return_by_index(string, middle_index) : return_by_range(string, middle_index - 1, middle_index)
```
hannazspot: I think of a bit easier and smaller solution like, what do you think about that 👇🏻 ?

```
def get_middle(string)
  middle_index = (string.size - 1) / 2.0
  string[middle_index.floor..middle_index.round]
end
```

hannazspot (approved):

PR #42, opened 2021-11-07
Skripchenkoveta hometask 2

skripchenkoVeta:


skripchenkoveta/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,26 @@
...
+  else
+    str[mid - 1] + str[mid]
+  end
+end
+
+def run_cli
+  loop do
+    print 'Enter string: '
+    str = gets.chomp
+    break if str == 'exit!'
```
jtethras: When you directly compare something to a _particular_ value, it's better to store this value somewhere in a const and compare with a const. This way, when you use this value multiple times in a project and at some point need to update this value, you won't have to search through your code and update it everywhere, but only update the const value.

jtethras (changes requested):
Overall looks good! But there is a potential for improvements. Please study following articles, improve whatever you find appropriate, and comment what you think :)

https://www.rubyguides.com/2018/10/puts-vs-print/
https://github.com/rubocop/ruby-style-guide#accessormutator-method-names (I know it's not a class here, but anyway I suggest to think on alternative naming)

Also, when you write your commit message, try to answer a question "What are these changes for?" and use present tense to answer the question. How would your rename your previous commit?

skripchenkoVeta:
> Overall looks good! But there is a potential for improvements. Please study following articles, improve whatever you find appropriate, and comment what you think :)
> 
> https://www.rubyguides.com/2018/10/puts-vs-print/ https://github.com/rubocop/ruby-style-guide#accessormutator-method-names (I know it's not a class here, but anyway I suggest to think on alternative naming)

The name of the method is not quite correct. It would be more correct to write 'midString' or 'middleString', right? Or just without 'get' , but then 'middle' name does not say what this method is about.

Also using 'p' instead of 'puts' (or 'print'), but it would be more correct to use 'puts' here. My code is already finished 'not debugging'

skripchenkoVeta:
> Also, when you write your commit message, try to answer a question "What are these changes for?" and use present tense to answer the question. How would your rename your previous commit?

Okay:)
I get it and will fix it

jtethras (approved):
🥇

PR #48, opened 2021-11-08
Silitski/hometask2

qqqwww123cdv:
Renamed the variable, used the ternary operator and changed the path.


qqqwww123cdv/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+def get_middle(str)
+  half = str.length / 2
+  str.length.even? ? str[(half - 1)..half] : str[half]
+end
+
+loop do
```
jekapulpit: also, you need this to be in a `run_cli` function

jekapulpit (changes requested):

PR #51, opened 2021-11-08
Alexey-Kren hometask-2

Alexey-Kren:


Alexey-Kren/hometask-2/get_middle_character.rb
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+def get_middle(str)
+  x = str.length / 2
+  str.length.odd? ? str[x] : str[x - 1] + str[x]
```
pavelbonda: You can use two arguments or range argument for `#[]`, for ex.:
```ruby
"some"[0, 2] 
=> "so"

"some"[0..1]
=> "so"
```

Alexey-Kren/hometask-2/get_middle_character.rb
```diff
@@ -0,0 +1,24 @@
...
+end
+
+def run_cli
+  loop do
+    p 'Enter string: '
```
pavelbonda: Consider moving texts to constants

pavelbonda:
Also, change PR name to include your surname

pavelbonda (approved):

PR #53, opened 2021-11-09
Gek62 hometask 2

GEK62:


Gek62/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env ruby
```
hrabiel, line 1: This is correct, but the file is still not executable, please fix it.

Gek62/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,19 @@
...
+
+# frozen_string_literal: true
+
+def get_middle(str)
+  str.length.even? ? str[str.length / 2 - 1] + str[str.length / 2] : str[str.length / 2]
+end
+
+def run_cli
+  loop do
+    puts 'Enter string: '
```
hrabiel: I think, the trailing space in the string is just redundant and can be removed.

hrabiel (changes requested):
As for the code implementation, good job! :+1: 

But there is a couple of problems, except those mentioned below:
1. Please, tell me you're not going to delete all these existing files :slightly_smiling_face: Something wrong happened when you were trying to pull the changes from the updated base branch.
2. There are already too many commits. It's not always important to have a clean commit history in feature branches, it depends on a team's agreement. But it's a good skill to have, so let's rewrite the commit history, please.
3. Another good piece of advice on commit messages. Please, try to make them as descriptive as possible. It's a good habit to develop. Also, there are different agreements, but the most common one is to use imperative mood, where the main verb is first. Like "Add login page", or "Fix admin permissions" etc.

hrabiel (approved):

PR #54, opened 2021-11-09
Natassia bon home work1

NastassiaBon:


Kaddyluck:
**Bad**:
1) Branch name - NastassiaBon:NatassiaBon-**HomeWork1**. Why? You need to create a new branch for every task
2) First commits from task 1, because of p.1

**Good**: 
Commit names with good descriptions!

PR #56, opened 2021-11-09
Mlshvch hometask 2

mlshvch:
I've completed Ruby-Lab hometask-2 with described requirements.


mlshvch/hometask-2/get_middle_character.rb
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env ruby
+
+# frozen_string_literal: true
+
+def get_middle(string)
+  len = string.length
```
StalemateInc, line 6: `len` is not the best variable name. Letters are free those days, so you might as well call it `length` or anything more descriptive than `len`.

mlshvch/hometask-2/get_middle_character.rb
```diff
@@ -0,0 +1,25 @@
...
+  mid = len / 2
```
StalemateInc, line 7: Same applies for `mid` and other obscure variable names.

mlshvch/hometask-2/get_middle_character.rb
StalemateInc: Also, useless variable assignment, this could easily be a part of the line below. You don't use the result of `len` in the scope of this method anywhere else.

mlshvch/hometask-2/get_middle_character.rb
```diff
@@ -0,0 +1,25 @@
...
+  len.even? == true ? string[mid - 1, mid] : string[mid]
```
StalemateInc, line 8: `if` conditions, in both full and ternary forms, by nature check if their body is equal to `true` or `false` to determine a code path. You don't have to write `... == true` manually, `if`/`unless` conditions will do it for you.

Please revise the following:
* how conditional statements work in Ruby (you can find the info in lecture materials)
* what types are treated as `true` and `false` in conditional statements (same as above, find in lectures).
* what return type is expected from method calls ending with a questionmark (spoiler alert - it's _usually_ true or false)

...and change the code accordingly.

mlshvch/hometask-2/get_middle_character.rb
```diff
@@ -0,0 +1,25 @@
...
+end
+
+def run_cli
+  loop do
+    puts 'Enter string: '
```
StalemateInc, line 13: Since you are using `puts` and read input starting from the next line, the extra space in the end of your string is useless.

mlshvch/hometask-2/get_middle_character.rb
StalemateInc: Also try to avoid magic numbers in future. Any numeric literal without an explicit purpose introduces confusion in future, so it's usually better to move them into constants or, for simple cases like this one, indicate the purpose with the variable name. Something like:
```ruby
middle_string_position = string.length / 2 # now it's more obvious why we are dividing by 2, we are looking for a middle!
```

mlshvch/hometask-2/get_middle_character.rb
```diff
@@ -0,0 +1,25 @@
...
+    input = gets.chomp
+    break if input == 'exit!'
+
+    if input.empty? == true
```
StalemateInc, line 17: The same comment about conditional statements also applies here.

mlshvch/hometask-2/get_middle_character.rb
```diff
@@ -0,0 +1,25 @@
...
+  len = string.length
+  mid = len / 2
+  len.even? == true ? string[mid - 1, mid] : string[mid]
+end
+
+def run_cli
+  loop do
+    puts 'Enter string: '
+    input = gets.chomp
+    break if input == 'exit!'
```
StalemateInc, line 15: It's usually a good practice to move non-interpolated strings (e.g. plain messages or some important strings like `exit!`) to constants to avoid reinstantiation and emphasize their importance. This way it's also much easier to find and change such strings later :)

mlshvch/hometask-2/get_middle_character.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    if input.empty? == true
+      puts 'String can not be blank!'
+    else
+      puts "Middle character: #{get_middle(input)}"
+    end
```
StalemateInc, line 21: Can probably be shortened to a ternary operation.

mlshvch/hometask-2/get_middle_character.rb
StalemateInc: Also, in idiomatic Ruby the usage of ranges is welcome when it comes to defining intervals, so getting a substring could be also written as:
```ruby
string[(mid - 1)..mid]
```

StalemateInc (changes requested):
Why does your PR contains both the first and the second home tasks? Please synchronize your fork and update the branch to have only the second one.

mlshvch/hometask-2/get_middle_character.rb
mlshvch: What practice is the best: to make only the first letter capital, or the whole constant?

mlshvch:
Will be better to  push changes as a new commit or refactor previous commits?

mlshvch/hometask-2/get_middle_character.rb
mlshvch: In case of ternary operation, length of line will be 94 characters, that is bigger than 80

mlshvch/hometask-2/get_middle_character.rb
StalemateInc: It depends on a context. Usually constants are named in all caps, names in camel case are reserved for classes and modules. Here is [a small excerpt from the book](https://www.oreilly.com/library/view/the-ruby-programming/9780596516178/ch04s03.html#:~:text=By%20convention%2C%20most%20constants%20are,letters%20and%20camel%20case%2C%20LikeThis%20.) which describes such conventions.

UPD: sorry, I broke the link, updated.

mlshvch/hometask-2/get_middle_character.rb
StalemateInc: All right, let's leave it as is to not to violate the code style.

StalemateInc:
> Will be better to push changes as a new commit or refactor previous commits?

Another commit with addressing the changes will be enough.

PR #57, opened 2021-11-09
hometask-2/create get middle

NastassiaBon:


hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,29 @@
...
+    puts 'Enter string: '
+    string = gets.chomp
+
+    break if string == 'exit'
+
+    if string.empty?
+      puts 'String can not be blank!'
+    else
+      puts "Middle character: #{get_middle(string)}"
+    end
```
Kaddyluck: Same with the ternary operator. Also, it can be a good style to extract pyre strings into constants. Mixed strings can be moved to the methods.

hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,29 @@
...
+
+def get_middle(string)
+  middle = string.length / 2 # your code here
+
+  if string.length.odd?
+    string[middle]
+  else
+    string[middle - 1] + string[middle]
+  end
+end
```
Kaddyluck: JFYI:
You can change that to ternary operator, like this (one line and do the same):
`string.length.odd? ? string[middle] : string[middle - 1] + string[middle]`

Also, you can use range args and `string[middle - 1] + string[middle]` become `string[(middle - 1)..middle]`

Kaddyluck:
Move your home task into your folder, pls

Kaddyluck:
Left some comments about the pretty code style above. Fixes are ​​welcome but not required

Kaddyluck (commented):
LGTM, but please, restore the first home task

PR #60, opened 2021-11-11
SLavahaim/hometask-2 

Slavahaim:
Add hometask-2 
Выполнил задание онлайн и же оформил, как того требовалось в hometask-2

### DIFF
--- a/Slavahaim/hometask-2/get_middle_charachter.rb
+++ b/Slavahaim/hometask-2/get_middle_charachter.rb
@@ -4,9 +4,9 @@
 EXIT_STATE = 'exit!'
 ERROR_MESSAGE = 'String can not be blank'
 def get_middle(inputed_string)
-  last_element = inputed_string.size - 1
+  premiddle_element = (inputed_string.size - 1) / 2
   middle_element = inputed_string.size / 2
-  inputed_string[last_element / 2..middle_element]
+  inputed_string[premiddle_element..middle_element]
 end
 
 def run_cli
--- a/Slavahaim/hometask-2/get_middle_charachter.rb
+++ b/Slavahaim/hometask-2/get_middle_charachter.rb
@@ -1,0 +1,21 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+def get_middle(str)
+  str[(str.size - 1) / 2..str.size / 2]
+end
+
+def run_cli
+  loop do
+    print 'Enter string: '
+    str = gets.strip
+    exit if str == 'exit!'
+    if str.empty?
+      puts 'String can not be blank!'
+    else
+      puts "Middle character: #{get_middle(str)}"
+    end
+  end
+end
+
+run_cli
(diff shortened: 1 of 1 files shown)


Slavahaim/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,21 @@
...
+
+def get_middle(str)
+  str[(str.size - 1) / 2..str.size / 2]
+end
+
+def run_cli
+  loop do
+    print 'Enter string: '
+    str = gets.strip
+    exit if str == 'exit!'
```
PFDeve: We can extract `exit!` to a constant for example `EXIT_STATE`
We can put this constant at the beginning of the script

Slavahaim/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,21 @@
...
+    if str.empty?
+      puts 'String can not be blank!'
```
PFDeve: The same. We can extract it to a constant like `ERROR_MESSAGE`

Slavahaim/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+def get_middle(str)
+  str[(str.size - 1) / 2..str.size / 2]
```
PFDeve: We can extract `str.size - 1` and `2..str.size / 2` into varibles and make a meaning of the varibles smth like `last_element` and `middle_element`. 
Try to avoid `magic number` like in `2..str.size / 2`. It's corrected by extracting into variables with meaning name. It makes a code more readable

PFDeve (changes requested):
Looks good to me :+1: 
There are notes which improve it

Slavahaim/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,21 @@
...
+end
+
+def run_cli
+  loop do
+    print 'Enter string: '
+    str = gets.strip
```
PFDeve: It will be readable if we name `str` like `inputed_string`

Slavahaim/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+EXIT_STATE = 'exit!'
+ERROR_MESSAGE = 'String can not be blank'
+def get_middle(inputed_string)
+  last_element = inputed_string.size - 1
+  middle_element = inputed_string.size / 2
+  inputed_string[last_element / 2..middle_element]
```
PFDeve: There is a magic numbar here. It is 2. I mean we can extract `2..middle_element` into the varible with `2..`

PFDeve: Let's rename `middle_element` in `middle_range`

PFDeve (approved):
👍

PR #62, opened 2021-11-11
Alhimick prog hometask 2

alhimick-prog:
Hometask-2 that requires create program, which finds the middle character(s) in string.

List of changes:

Added:
+ string constant for all messages and for string for exit the program

Deleted:
- extra method `.to_s`
- extra method `validation`
- extra variables

Rework:
* loop implementation from `unless` to `loop`
* control flow from `case` to `if/else`


alhimick-prog/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,33 @@
...
+  if string.length.odd?
+    string[mid]
+  else
+    string[mid - 1] + string[mid]
+  end
+end
+
+def validation(string)
+  if string.empty?
+    puts 'String can not be blank!'
```
lincAlinc: @alhimick-prog what do you think about adding constant for this string?

alhimick-prog/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,33 @@
...
+  else
+    true
+  end
+end
+
+def run_cli
+  string = ''
+  until string == 'exit!'
+    stop = false
+    puts 'Please, enter string:'
```
lincAlinc: @alhimick-prog what do you think about adding constant for this string?

alhimick-prog/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,33 @@
...
+    string = gets.chomp.to_s
```
lincAlinc: @alhimick-prog why do you add `to_s` method there?

alhimick-prog/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,33 @@
...
+def get_middle(string)
+  mid = string.length / 2
+  if string.length.odd?
+    string[mid]
+  else
+    string[mid - 1] + string[mid]
+  end
+end
+
+def validation(string)
```
lincAlinc: @alhimick-prog your method return `true` or `false`. In this case method name needs to end with `?`. For example `valid?`

alhimick-prog/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+end
+
+def run_cli
+  string = ''
+  until string == 'exit!'
+    stop = false
+    puts 'Please, enter string:'
+    string = gets.chomp.to_s
+    stop = true if string == 'exit!'
```
lincAlinc: @alhimick-prog `string == 'exit!'` return `true` or `false`. Do we really need `true if` and `stop = false`?

alhimick-prog/hometask-2/get_middle_charachter.rb
```diff
@@ -0,0 +1,33 @@
...
+    puts "Middle character(s): #{get_middle(string)}" if validation(string) && !stop
```
lincAlinc: @alhimick-prog Can we add constant for `"Middle character(s):"`?

lincAlinc (changes requested):

lincAlinc (changes requested):
Could you try to rewrite `run_cli` with `loop`?

alhimick-prog/hometask-2/get_middle_charachter.rb
alhimick-prog: Yes, we can. I added constants for all messages.

alhimick-prog/hometask-2/get_middle_charachter.rb
alhimick-prog: Think good. I added constants for all messages.

alhimick-prog/hometask-2/get_middle_charachter.rb
alhimick-prog: Think good. I added constants for all messages.

alhimick-prog/hometask-2/get_middle_charachter.rb
alhimick-prog: I thought it's needed for correct work with user input, but program works correct without this.

lincAlinc: @alhimick-prog `gets` get back a string. when you use chomp for string - it returns a string.

alhimick-prog: Yes, this is a lack of knowledge about the method `gets`

lincAlinc (approved):
Cool!

PR #63, opened 2021-11-11
trueananas17/hometask-2

prakofyeu:


kukharava (approved):
LGTM

One notice: 

It would be nice to make the commit names more meaningful.
Please write in the commit message what exactly you did, so that another developer can read this message and understand which changes have been made

PR #64, opened 2021-11-11
AndrewTehan-hometask-3

AndrewTehan:
[hometask-3](https://github.com/AndrewTehan/Ruby-Lab#%D0%B7%D0%B0%D0%B4%D0%B0%D0%BD%D0%B8%D0%B5-3)


hannazspot (commented):
1. Let's remove the `codewars.rb` file: it seems useless. Moreover, you copy-pasted code from codewars to rainfall, so that does not make any sense to me 🤷🏻‍♀️ 
2. let's split code in `rainfall.rb` by classes and move them to separate files. It will make code readable and more sensible

AndrewTehan/hometask-3/rainfall_statistics.rb
```diff
@@ -0,0 +1,51 @@
...
+    extract_statistic_values.map { |town_monthly_rainfall| town_monthly_rainfall - mean }
+  end
+
+  def average(values)
+    values.sum(&:to_f) / values.size
+  end
+
+  def calculate_variance(values)
+    values.sum { |value| value**2 } / values.size
+  end
```
hannazspot: I'm wondering why methods are called according to different concepts. They both do some calculations, but one of them is called `calculate_*` and another one is just `average`

So, it's not so critical there, but it's better to start to control such stuff even in small programs.

hannazspot: Moreover, they seem to me as methods that don't belong to this Class.

- you wrote them in such a way that they can be reused in other contexts 
- you called them in such a way that I don't even know that they're a part of `RainfallStatistics`
- they are independent pieces of code 

So, I really like the way that you wrote that part of the code ☝🏻  but I'd move it to a separate module, like a tool, that helps `RainfallStatistics` handle its purposes and responsibilities. It will make a class more clearer + it will make the calculation part reusable in any other part of a program

AndrewTehan/hometask-3/rainfall_statistics.rb
```diff
@@ -0,0 +1,51 @@
...
+  def variance
+    return -1 if missing_data?
+
+    calculate_variance(calculate_mean_differences)
+  end
+
+  private
+
+  def extract_statistic_values
+    @extract_statistic_values ||= find_town_statistics.scan(/[0-9]*\.[0-9]*/).map(&:to_f)
```
hannazspot: You're lucky that you have an easy regular expression like `/[0-9]*\.[0-9]*/` and I could guess what it does. But sometimes it's very hard to get a purpose of regexp.

I'd recommend avoiding using NONAME stuff, especially when it could not be clear what it's for. Let's move it to a constant 💃🏻

hannazspot: Please have a look at https://rubular.com/ It will help you use simple and elegant concepts of writing regular expressions. Instead of writing `[0-9]`, you can replace it with `\d`. What do you think about it? is it nicer a bit, nuh? 😎😉✌🏻

Screen Shot 2021-11-15 at 1 00 02 PM
Screen Shot 2021-11-15 at 12 59 41 PM

AndrewTehan/hometask-3/cli.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+ENTER_MESSAGE = 'Enter city name: '
+ERROR_MESSAGE = 'City name can not be blank!'
+EXIT_MESSAGE = 'exit!'
```
hannazspot: I disagree with constants' scope. I'd move it to Cli class since it's used there.

A small note for the future: I'd recommend you avoid globalizing stuff since it's a bad practice.

AndrewTehan/hometask-3/cli.rb
```diff
@@ -0,0 +1,44 @@
...
+    loop do
+      input = wait_for_input
+
+      exit if input == EXIT_MESSAGE
+
+      if input.empty?
+        show_error
+      else
+        show_result(input)
+      end
```
hannazspot: Why don't you use ternary operator `input.empty? ? show_error : show_result(input)`?

it looks lighter than if-else construction

AndrewTehan/hometask-3/cli.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+  end
+
+  private
+
+  def data
+    @data ||= File.open('data').read
+  end
```
hannazspot: It's not critical, but it's better to consider an order of methods' calling and their place in a file

you call `data` only in `show_result`. For some reason, `data` is placed above the method called it in code. So it's highly recommended to take into account such micro details in programming. Jumping through the codebase is distrusting and not so elegant, as Ruby code is supposed to be.

AndrewTehan/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require './cli'
+require './rainfall_statistics'
+
+def run_cli
+  Cli.new.run_cli
```
hannazspot: `Cli.new.run` is a quite readable naming, `Cli.new.run_CLI` is redundant a bit to me

AndrewTehan/hometask-3/rainfall_statistics.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+# this class is used for calculate mean and variance values for town rainfall
+class RainfallStatistics
+  attr_reader :town, :data
```
hannazspot: I'd move it to `private` scope in this class, cause these readers are not used out of class

AndrewTehan/hometask-3/rainfall_statistics.rb
```diff
@@ -0,0 +1,51 @@
...
+    return -1 if missing_data?
+
+    average(extract_statistic_values)
+  end
+
+  def variance
+    return -1 if missing_data?
+
+    calculate_variance(calculate_mean_differences)
+  end
```
hannazspot: You could гыу a block there to avoid writing the same piece of code. It's not a big deal there, but ... it'll be cooler if you use some Ruby possibilities where it's suitable to apply 💁🏼

```
def mean
  town_exists? { average(extract_statistic_values) }
end

def mean
  town_exists? { calculate_variance(calculate_mean_differences) }
end

def town_exists?(&calculations)
  find_town_statistics.present? ? calculation.call : return -1
end
```

Does it look better a bit? 👀

hannazspot (approved):
Well done 👏🏻 😎

PR #65, opened 2021-11-11
TehanovAnton/hometask-3

TehanovAnton:


TehanovAnton/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+def run_cli
+  RainfallCli.new.run_cli
+end
+
+# RainfallCli designed to initialise data for run_cli method
```
dmokreckiy: Wouldn't it be better if all of these files where separated?
For example:
1. `rainfall.rb` will be execution file
2. `RainfallCli` will be a separate file.
3. `RainfallStatistics` another separate file.

TehanovAnton/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,91 @@
...
+class RainfallCli
+  def run_cli
+    loop do
+      puts 'Enter city name:'
+      input = gets.chomp
+
+      break if input == 'exit!'
+
+      input.empty? ? show_error_message : show_result(input)
```
dmokreckiy: Can you apply [guard clause](https://docs.rubocop.org/rubocop/1.0/cops_style.html#styleguardclause) here?

TehanovAnton/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+  end
+
+  private
+
+  def show_error_message
+    puts "City name can't be blank"
+  end
+
+  def show_result(input)
```
dmokreckiy: `input` is not a good parameter name, don't you think?
Try to do a better naming keeping in mind that we always query some city.

TehanovAnton/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,91 @@
...
+    statistics = RainfallStatistics.new(input, data)
+    puts "Rainfall mean: #{statistics.mean}"
+    puts "Rainfall variance: #{statistics.variance}"
+  end
+
+  def data
+    @data ||= File.open('rainfall_data').read
```
dmokreckiy: Codewars mentioned usage of two file. Can you rewrite this class to accept file name and use it a variable instead of fixed value?

TehanovAnton/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,91 @@
...
+      town_rainfall_data.start_with?("#{town}:")
+    end
+  end
+
+  def missing_data?
+    !find_town_rainfall_data
+  end
+
+  def parse_rainfall_values
+    @parse_rainfall_values ||= find_town_rainfall_data.scan(/\d+.\d+/).map(&:to_f)
```
dmokreckiy: What do you think about naming it `rainfall_values`? Check all places of this method invocation and see if this change can make code more readable.

TehanovAnton/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  def average(values)
+    values.sum / values.size
+  end
+
+  def calculate_rainfall_diviations
```
dmokreckiy: Same here. Try it without `calculate_`

TehanovAnton/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,91 @@
...
+    parse_rainfall_values.map { |value| value - mean }
+  end
+
+  def calculate_variance(rainfall_diviations)
```
dmokreckiy: Same here. Try it without `calculate_`

dmokreckiy (changes requested):
@TehanovAnton check out my comments
Required changes:
1) Split into separate files.
2) Apply guard clause rule.
3) Rename parameter.
Optional:
1) Add ability to supply data filename.
2) Rename methods if you see it fit.

TehanovAnton/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require './rainfall_cli'
```
dmokreckiy: Please rewrite without `./`. There are a special require for that.

TehanovAnton/hometask-3/rainfall_cli.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+require './rainfall_statistics'
```
dmokreckiy: Same as above

TehanovAnton/hometask-3/rainfall_cli.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def run_cli
+    loop do
+      puts 'Enter city name:'
+      input = gets.chomp
+
+      break if input == 'exit!'
+
+      show_error_message if input.empty?
+      show_result(input)
```
dmokreckiy, line 20: Usually there is an empty line after guard clause but it will do. :+1:

TehanovAnton/hometask-3/rainfall_cli.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+require './rainfall_statistics'
+
+# RainfallCli designed to initialize data for run_cli method
+class RainfallCli
+  def initialize(data_file)
+    @data_file = data_file
```
dmokreckiy, line 8: I would say `data_file_name` but it will do.

TehanovAnton/hometask-3/rainfall_statistics.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+# RainfallStatistics designed to calculate mean and variance values
+class RainfallStatistics
+  attr_reader :town, :data
+
+  def initialize(town, data)
```
dmokreckiy: Just a sidenote - in RainfallCli `city`, here `town`. Try to be consistent.

dmokreckiy (changes requested):
Please fix `require './xxx`

dmokreckiy (approved):
LGTM

PR #69, opened 2021-11-12
Matveibos-hometask-2

Matveibos:
Домашняя работа 2


Matveibos/hometask-2/get_middle_character.rb
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env ruby
+
+# frozen_string_literal: true
+
+def run_cli
+  loop do
+    puts 'Enter string'
```
vol6tmetr: It is better to define strings and numbers used in application as a constant variables. You can find more information about this practice if you search "magic strings/numbers"

Matveibos/hometask-2/get_middle_character.rb
```diff
@@ -0,0 +1,28 @@
...
+    string = gets.chomp
+    break if string == 'exit!'
+
+    if string.empty?
```
vol6tmetr: It is completely fine to use if/else here, but check out how you can rewrite this statement with the ternary expression. You can use them, if it is not affecting readability for the code, and this is the case for this code, IMHO

Matveibos/hometask-2/get_middle_character.rb
```diff
@@ -0,0 +1,28 @@
...
+      puts 'String can not be blank!'
+    else
+      puts "Middle character: #{get_middle(string)}"
+
```
vol6tmetr, line 9: Extra empty line (not needed)

Matveibos/hometask-2/get_middle_character.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+  end
+end
+
+def get_middle(str)
+  if str.length.even?
```
vol6tmetr: Again, I think that ternary expression will work here as well

Matveibos/hometask-2/get_middle_character.rb
```diff
@@ -0,0 +1,28 @@
...
+    puts str[str.length / 2 - 1] + str[str.length / 2]
```
vol6tmetr: Check this out, you use `str.length / 2` 3 times here, I think it is better to move it to the variable, if you use this so many times
Also you can use ranges to iterate with the strings in Ruby, check how you can rewrite `get_middle` function with the ranges

Matveibos/hometask-2/get_middle_character.rb
Matveibos: Do you mean combine break if string == 'exit!'  and if string.empty? ...  with the ternary expression?

vol6tmetr: No, I mean that you can rewrite that to this:
```
string.empty? ? puts 'String can not be blank!' : puts "Middle character: #{get_middle(string)}"
```

Matveibos: I did it like this :
if st.empty? ? (puts 'String can not be blank!') : (puts RESULT + get_middle(st))
# comment, but rubocop warns: Lint/EmptyConditionalBody: Avoid if branches without a body.
end

Matveibos/hometask-2/get_middle_character.rb
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env ruby
+
+# frozen_string_literal: true
+
+ENTER_STRING = 'enter string'
+ERROR = 'String can not be blank!'
+EXIT = 'exit!'
+RESULT = 'Middle character:'
```
vol6tmetr, line 8: Add empty line between constants and other parts of code

Matveibos/hometask-2/get_middle_character.rb
```diff
@@ -0,0 +1,23 @@
...
+  str[slen.floor..slen.ceil]
+end
+
+def run_cli
+  loop do
+    puts ENTER_STRING
+    st = gets.chomp
+    break if st == EXIT
+
+    st.empty? ? (puts 'String can not be blank!') : (puts RESULT + get_middle(st))
```
vol6tmetr: You just moved `'String can not be blank!'` to the `ERROR` constant, so lets use it there
Also prefer string interpolation, rather then concatenation, at least in this case: `RESULT + get_middle(st)`

Matveibos: please excuse my inattention)
Ok!

vol6tmetr (approved):

PR #75, opened 2021-11-14
Valeriyapetrova/hometask 3

ghost:


valeriyapetrova/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,68 @@
...
+      puts "Rainfall mean: #{mean}"
+      puts "Rainfall variance: #{variance}"
+    end
+
+    repeat
+  end
+
+  def take_data
+    data = File.new('take_data.txt', 'r:UTF-8')
+    @data = data.read
```
TanyaAkulich:
```suggestion
    @data = File.new('take_data.txt', 'r:UTF-8').read
```

btw why do you need 'r:UTF-8' herer?

valeriyapetrova/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  def repeat
+    @user_input = user_input
+
+    run_cli
+  end
+
+  def mean
+    take_data
+    whole_string = @data.split("\n")
```
TanyaAkulich: let's move `"\n"` to a const (smith like `TEXT_SEPARATOR`

valeriyapetrova/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,68 @@
...
+      puts 'Mistake!!!'
+    else
+      puts "Rainfall mean: #{mean}"
+      puts "Rainfall variance: #{variance}"
+    end
+
+    repeat
+  end
+
+  def take_data
```
TanyaAkulich: move this method to the private scope

valeriyapetrova/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    run_cli
+  end
+
+  def mean
+    take_data
+    whole_string = @data.split("\n")
+    whole_string.each do |string|
+      city = /\w+:/.match(string).to_s.delete(':')
+      rainfall = string.scan(/([0-9-]+\.[0-9])/).flatten.map(&:to_f)
```
TanyaAkulich: regular expressions also to consts

valeriyapetrova/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,68 @@
...
+      @city_and_values[city] = rainfall
+    end
+
+    return -1 unless @city_and_values[@user_input]
+
+    @city_and_values[@user_input].sum / 12
```
TanyaAkulich: 12 to const (or use `@city_and_values[@user_input].size` instead bc w/o context it's a magic number)

valeriyapetrova/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,68 @@
...
+  def repeat
+    @user_input = user_input
+
+    run_cli
+  end
+
+  def mean
+    take_data
+    whole_string = @data.split("\n")
+    whole_string.each do |string|
```
TanyaAkulich: let's move this iterator to a new method and will call it from `initialize` to identify `@city_and_values` at the beginning

TanyaAkulich (changes requested):
you can implement a new `current_city` value instead of using @city_and_values[@user_input]

and .txt for your example file is not required in that case

TanyaAkulich (approved):

PR #79, opened 2021-11-14
Alhimick-prog/hometask-3

alhimick-prog:
Hometask-3 that requires create program, which finds the desired string, selects numerical values from it and calculates the average value and valuates of values.

This version contains three files: two files with input data and the file of the program itself. All the necessary actions for search and mathematical calculations have been implemented.

Program is started and used in accordance with technical task.

List of changes 1:
Added:
+ comments for methods

Deleted:

Rework:
* method `variance`: work with array with `.inject` method

Fixes:
* constant names (MOUNTH_NUMBER -> MONTH_NUMBER, ERROR_MOUNT_NUMBER -> ERROR_MONTH_NUMBER)

List of changes 2:
Added:

Deleted:
- comments for methods

Rework:
* many methods. Now program does not create a substring from the data to create an array of numbers. Instead, numbers are pulled directly from the data.
* renamed constant TOWN_ERROR to TOWN_SEARCH_ERROR_NUMBER

Fixes:


alhimick-prog/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,101 @@
...
+
+INPUT_PROMPT = 'Enter city name:'
+ERROR_BLANK_STRING = 'City name can not be blank!'
+ERROR_FILE_READ_FAIL = 'Error: data file not found'
+ERROR_MOUNT_NUMBER = 'Error: something went wrong when counting months.'
+MEAN = 'Rainfall mean:'
+VARIANCE = 'Rainfall variance:'
+EXIT_STRING = 'exit!'
+TOWN_ERROR = -1
+MOUNTH_NUMBER = 12
```
lincAlinc: @alhimick-prog do you mean `MONTH`?

alhimick-prog/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+INPUT_PROMPT = 'Enter city name:'
+ERROR_BLANK_STRING = 'City name can not be blank!'
+ERROR_FILE_READ_FAIL = 'Error: data file not found'
+ERROR_MOUNT_NUMBER = 'Error: something went wrong when counting months.'
```
lincAlinc: @alhimick-prog `ERROR_MONTH_NUMBER`?

alhimick-prog/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,101 @@
...
+MEAN = 'Rainfall mean:'
+VARIANCE = 'Rainfall variance:'
+EXIT_STRING = 'exit!'
+TOWN_ERROR = -1
```
lincAlinc: @alhimick-prog what do you think about `TOWN_ERROR_NUMBER`?

alhimick-prog/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,101 @@
...
+  array = form_array(town, strng)
+  return TOWN_ERROR unless array
+
+  mean_val = mean(town, strng)
+  i = 0
+  sum = 0
+  while i < array.size
+    sum += (array[i] - mean_val)**2
+    i += 1
+  end
```
lincAlinc: @alhimick-prog could you try to rewrite this with ruby style? A little hint: you can use `reduce` or `inject` method

alhimick-prog/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,101 @@
...
+    false
+  end
+end
+
+def mean(town, strng)
+  array = form_array(town, strng)
+  return TOWN_ERROR unless array
+
+  puts ERROR_MOUNT_NUMBER if array.size != MOUNTH_NUMBER
+  array.sum / array.size
```
lincAlinc: @alhimick-prog you print error but continue to calculate `mean`. I think we need to stop the program with error, right?

alhimick-prog/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,101 @@
...
+end
+
+def variance(town, strng)
+  array = form_array(town, strng)
+  return TOWN_ERROR unless array
```
lincAlinc: @alhimick-prog duplicate from `mean` method. How we can improve this?

alhimick-prog/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,101 @@
...
+MOUNTH_NUMBER = 12
+CURRENT_FILE = 'data.txt'
+
+def file_reader(filename)
+  if File.exist?(filename)
+    file = File.new(filename, 'r')
+    file.sysread(file.size)
+    # I know i could use a method .readlines to get an array from file,
+    # but in the condition of assignment
+    # data is passed to methods 'mean' and 'variance' as a string.
```
lincAlinc, line 21: @alhimick-prog Why do you decide to send data to `mean` and `variance` as a string? 
The purpose of these methods is just to calculate `mean` and `variance`, you need to pass the data and don't think about how to process strings

alhimick-prog/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,101 @@
...
+
+def form_array(town, strng)
+  if town_pos(town, strng)
+    start_search_pos = town_pos(town, strng) + town.size + 1
+    substr = form_substring(start_search_pos, strng)
+    substr.scan(/\d+\.*\d*/).map(&:to_f)
+  else
+    false
+  end
+end
```
lincAlinc: @alhimick-prog Could you please describe in detail what are you doing in these methods?(`town_pos`, `form_substring`, `form_array`)

lincAlinc (changes requested):

alhimick-prog/hometask-3/rainfall.rb
alhimick-prog: @lincAlinc yes, i mean `MONTH`

alhimick-prog/hometask-3/rainfall.rb
alhimick-prog: @lincAlinc yes, `ERROR_MONTH_NUMBER`

alhimick-prog/hometask-3/rainfall.rb
alhimick-prog: @lincAlinc in the condition of the problem on the codewars:
"if functions `mean` or `variance` have as parameter `town` a city which has no records return `-1` or `-1.0` (depending on the language)"
In this constant I store the value that needs to be returned in accordance with the condition of the task.
Maybe it makes sense to rename to `MESSAGE_ERROR_TOWN_SEARCH`?

alhimick-prog/hometask-3/rainfall.rb
alhimick-prog: @lincAlinc thank for the hint. It looks better now.

alhimick-prog/hometask-3/rainfall.rb
alhimick-prog: @lincAlinc in this task, this check of quantity is not required at all, since everything works correctly and with verified data. I added this check in case of working with other data sources that may be incorrectly processed or may themselves be incorrect (for example, the line contains data not for all months). In this case, the user will receive a warning, but the program will continue to run. Perhaps this check is not needed?

alhimick-prog/hometask-3/rainfall.rb
alhimick-prog: @lincAlinc if I completely follow the conditions of the task and pass in methods `mean` and `varince` only `town` and `strng`, then I do not know how to avoid this repetition. If we deviate from the strict rules, I could pass an additional variable. Also I can put both methods in one class and store the calculated value in an instance variable.

alhimick-prog/hometask-3/rainfall.rb
alhimick-prog: @lincAlinc "function: `mean(town, strng)` should return the average of rainfall for the city `town` and the `strng` `data `or `data1`"
On the site with the task, `data` and `data1` are strings. So in my solution the data is also passed as a string.

alhimick-prog/hometask-3/rainfall.rb
alhimick-prog: @lincAlinc OK. I think this is a good idea

alhimick-prog/hometask-3/rainfall.rb
lincAlinc: @alhimick-prog -1 - it's number, you store this as integer. I think `TOWN_SEARCH_ERROR_NUMBER`

alhimick-prog/hometask-3/rainfall.rb
lincAlinc: @alhimick-prog ok

alhimick-prog/hometask-3/rainfall.rb
lincAlinc: @alhimick-prog if you want just add warning, you need to rename this `ERROR_MOUNT_NUMBER`

alhimick-prog/hometask-3/rainfall.rb
jekapulpit: how about `TOWN_SEARCH_ERROR_CODE` ? 😄

ramantechart (changes requested):
This PR contains commit from hometask-1 and hometask-2

PR #83, opened 2021-11-15
Mlshvch hometask 3

mlshvch:
I've completed hometask 3 with all describes requirements.


mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env ruby
+
+# frozen_string_literal: true
+
+EXIT_COMMAND = 'exit!'
+ERROR_MESSAGE = 'City name can not be blank!'
+
+def extract_data(strng)
```
StalemateInc, line 8: Does making a variable name one symbol shorter helps in any way? Let's use the names that are descriptive enough and can be easily readable. `strng` is neither of those. It does not speak about what kind of string it is or what information does it contain.

StalemateInc: From this point on, let's use variable names that: a) easy to read (Ruby is usually advertised as a language, close to plain English); b) descriptive. `strng` is neither of those two. This name is shortened for no reason and does not even indicate the purpose of a variable or even its contents.

I see that it's the same name as in requirements, but please fix it anyway. In Russian version of the task there is nothing about keeping variable names the same.

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,55 @@
...
+  hash = {}
+  strng.split("\n").each do |data|
+    array = []
+    city, values = data.split(':')
+    results = values.split(',')
+    results.each { |value| array.push(value.split(' ')[1]) }
+    hash[city] = array
+  end
+  hash
```
StalemateInc: If you need to do any kind of sequential processing and aggregate results in a different object, please consider using [Enumerable#each_with_object](https://devdocs.io/ruby~3/enumerable#method-i-each_with_object) or [Enumerator#with_object](https://devdocs.io/ruby~3/enumerator#method-i-with_object). This way you won't have to declare stuff like `hash = {}`, where the initial state of the hash is never used and variable exists outside of enumeration process. It will also help to eliminate call to `hash` in the end :)

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,55 @@
...
+end
+
+def mean(town, strng)
+  mean = -1.0
+  sum = nil.to_f
```
StalemateInc: What's the point of such conversion? Why can't you simply use `0.0`?

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,55 @@
...
+    city, values = data.split(':')
+    results = values.split(',')
+    results.each { |value| array.push(value.split(' ')[1]) }
+    hash[city] = array
+  end
+  hash
+end
+
+def mean(town, strng)
+  mean = -1.0
```
StalemateInc: I already mentioned magic numbers in your previous PR, such things should be moved to constants.

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,55 @@
...
+EXIT_COMMAND = 'exit!'
+ERROR_MESSAGE = 'City name can not be blank!'
+
+def extract_data(strng)
+  hash = {}
+  strng.split("\n").each do |data|
+    array = []
+    city, values = data.split(':')
+    results = values.split(',')
+    results.each { |value| array.push(value.split(' ')[1]) }
```
StalemateInc: In case with your `results` variable you can simply chain the calls of the methods, the actual values from `results` are not used anywhere else in the method. Also, [Enumerable#map](https://devdocs.io/ruby~3/enumerable#method-i-map) does everything you want and in much more elegant way.

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,55 @@
...
+# frozen_string_literal: true
+
+EXIT_COMMAND = 'exit!'
+ERROR_MESSAGE = 'City name can not be blank!'
+
+def extract_data(strng)
+  hash = {}
+  strng.split("\n").each do |data|
+    array = []
+    city, values = data.split(':')
```
StalemateInc: Same comment about variable naming - `array`, `values`, ... - don't not indicate anything useful.

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+  hash
+end
+
+def mean(town, strng)
+  mean = -1.0
+  sum = nil.to_f
+  data = extract_data(strng)
+  unless data[town].nil?
+    data[town].each { |value| sum += value.to_f }
```
StalemateInc: How do you think, what is more optimal: convert strings to floats each time you pull the data for a new city or do it once during the initial data processing?

StalemateInc: Also, usage of [Enumerable#sum](https://devdocs.io/ruby~3/enumerable#method-i-sum) is highly recommended, that's a Ruby way to sum a collection of numbers. You can even specify a starting value there, very convenient.

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,55 @@
...
+    hash[city] = array
+  end
+  hash
+end
+
+def mean(town, strng)
+  mean = -1.0
+  sum = nil.to_f
+  data = extract_data(strng)
+  unless data[town].nil?
```
StalemateInc: Falling back to once somewhere initialized value is not that great idea, at least not in this case. It's much better to not to outsmart yourself and simply return early with a constant if we went unhappy path. Something like:
```ruby
return DEFAULT_MEAN unless data[town]
```

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,55 @@
...
+    data[town].each { |value| sum += value.to_f }
+    mean = sum / data[town].length
```
StalemateInc: Does every city has it's own amount of months in a year?

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,55 @@
...
+
+def run_cli
+  loop do
+    puts 'Enter city name:'
+    input = gets.chomp
+    break if input == EXIT_COMMAND
+
+    success_message = "Rainfall mean: #{mean(input, '')}\nRainfall variance: #{variance(input, '')}"
+
+    puts input.empty? ? ERROR_MESSAGE : success_message
```
StalemateInc: The case when using ternary operator makes more harm than good. You could get away with usual "if..else" statement and not glue strings together with "\n".

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,55 @@
...
+  variance
+end
+
+def run_cli
+  loop do
+    puts 'Enter city name:'
+    input = gets.chomp
+    break if input == EXIT_COMMAND
+
+    success_message = "Rainfall mean: #{mean(input, '')}\nRainfall variance: #{variance(input, '')}"
```
StalemateInc: You can barely get anything out of an empty string, that's pointless.

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,55 @@
...
+    results.each { |value| array.push(value.split(' ')[1]) }
+    hash[city] = array
+  end
+  hash
+end
+
+def mean(town, strng)
+  mean = -1.0
+  sum = nil.to_f
+  data = extract_data(strng)
```
StalemateInc: So, as far as I can see, you are extracting the same data each time you request call mean and variance. It's not optimal, much better option is to extract the data once before calling those methods and then work with already extracted data, don't you think?

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,55 @@
...
+  variance = -1.0
+  sum = nil.to_f
+  data = extract_data(strng)
+  unless data[town].nil?
+    mean_value = mean(town, strng)
+    data[town].each { |value| sum += (value.to_f - mean_value)**2 }
+    variance = sum / data[town].length
+  end
+  variance
+end
```
StalemateInc: Every comment above applies here as well.

StalemateInc (changes requested):
It doesn't even work. Where is the data?
The program always returns `-1.0` for both mean and variance.

mlshvch/hometask-3/rainfall.rb
mlshvch: If you mean the 48th string, this is necessary due to Rubocop recommendations (Layout/EmptyLineAfterGuardClause)

StalemateInc: No, I meant the second argument for `mean` and `variance` methods, which is supposed to hold your data about cities and rainfall metrics.

mlshvch:
Many things are done the way they are because of Codewars Task specification

mlshvch/hometask-3/rainfall.rb
mlshvch: I don't understand to to refactor my code using each_with_object

StalemateInc:
> Many things are done the way they are because of Codewars Task specification

Yes, I understand that you've relied on the original Codewars task, but you have additional specifications in your README file as well. Your requirement was not only to solve a Codewars task, but also modify it to satisfy the requirements from the README and right now your solution doesn't even function the proper way. As you could notice, on yours or mine computer there is no special console with prefilled data to use inside the program. You can hardcode the rainfall metrics as a constant (thought I strongly recommend not to do it) or add it as a file and read from it during the program execution.

The wording in a task could be an oversight from the mentors group, but in any case, let's be flexible and adapt the solution to what's described in README.

mlshvch/hometask-3/rainfall.rb
StalemateInc: I'll give you an example:
```suggestion
  strng.split("\n").each_with_object({}) do |data, city_rainfall_stats|
    array = []
    city, values = data.split(':')
    results = values.split(',')
    results.each { |value| array.push(value.split(' ')[1]) }
    city_rainfall_stats[city] = array
  end
```
What we achieved with such a small change:
1. no out-of-scope variable declaration -> more Rubyish way to do things;
2. `each_with_object` automatically returns the object used as accumulator, in this case it's `city_rainfall_stats`, as a result you don't have to explicitly return it in the end;
3. it's more laconic 🙂

Did it get more clear now? If not, then we can setup a call and talk about it.
Addressing the rest of the comments is up to you, I did nothing else other than demonstrated how you could use `each_with_object` in this particular case.

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,52 @@
...
+def run_cli
+  loop do
+    puts 'Enter city name:'
+    input = gets.chomp
+    break if input == EXIT_COMMAND
+
+    if input.empty?
+      puts ERROR_MESSAGE
+    else
+      puts "Rainfall mean: #{mean(input, @rainfall_data)}"
```
StalemateInc, line 46: Better than it was before, but still not the best approach. You see, you've just created an instance variable `@rainfall_data`, but you don't have any other enclosing context to isolate it. Your Ruby code is currently written in procedural style, which means that it is executed in a scope of `main` object - the topmost object, where all the other defined methods, constants and the rest reside. Since it's a global topmost object, any instance variable will live there until the execution of Ruby process stops and can possibly override any other existing instance variables with the same name in `main` object itself. In other words, this approach has the almost the same downsides as using global variables and highly discouraged to be used in `main` scopes. If you don't have any modules or classes to which this instance variable could belong, then it's better to not use it.

The whole lifecycle of your program is within one method, `run_cli`. What's better to do here is to read the contents once before the beginning of the `loop` statement and then just pass the local to `mean` / `variance` the usual way.

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env ruby
+
+# frozen_string_literal: true
+
+file = File.open('data.txt', 'r')
+data = file.read
+file.close
```
StalemateInc, line 7: Relative paths does not play good when you are trying to start your script not from the exact directory where it is located. For example, it does not work, when I try to run it from `mlshvch/` folder:
```bash
> pwd
Ruby-Lab/mlshvch
> hometask-3/rainfall.rb
Traceback (most recent call last):
	2: from hometask-3/rainfall.rb:5:in `<main>'
	1: from hometask-3/rainfall.rb:5:in `open'
hometask-3/rainfall.rb:5:in `initialize': No such file or directory @ rb_sysopen - data.txt (Errno::ENOENT)
```

A possible solution there is to use the possibilities of `File` Ruby class and `__dir__` / `__FILE__` constants to build up a correct absolute path to a file.

StalemateInc: Also, the block notation of `File.open` closes file automatically at the end of the block. If you use that you save your time and don't worry about file handlers. As another side note: `File#read` does everything for you: opens a file, reads the contents and closes it afterwards. Take a look ;)

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,52 @@
...
+
+EXIT_COMMAND = 'exit!'
+ERROR_MESSAGE = 'City name can not be blank!'
+DEFAULT_VALUE = -1.0
```
StalemateInc, line 11: Default value of what?

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,52 @@
...
+  return DEFAULT_VALUE unless data[town]
+
+  data[town].sum(&:to_f) / 12
+end
+
+def variance(town, data)
+  return DEFAULT_VALUE unless data[town]
+
+  mean_value = mean(town, @rainfall_data)
+  data[town].sum { |value| (value.to_f - mean_value)**2 } / 12
```
StalemateInc, line 34: `12` is still a magic number.

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,52 @@
...
+
+EXIT_COMMAND = 'exit!'
+ERROR_MESSAGE = 'City name can not be blank!'
+DEFAULT_VALUE = -1.0
+
+def extract_data(data_string)
+  data_string.split("\n").each_with_object({}) do |data, city_rainfall_stats|
+    town, rainfall_stats = data.split(':')
+    city_rainfall_stats[town] = rainfall_stats.split(',').each_with_object([]) do |results, month_rainfall_stats|
+      month_rainfall_stats << results.split(' ')[1]
```
StalemateInc, line 17: If you need to iterate through collection, performing transformation over the values, and get another collection as a result, it is much easier to use [Enumerable#map](https://devdocs.io/ruby~3/enumerable#method-i-map), like I mentioned before.

```ruby
city_rainfall_stats[town] = rainfall_stats.split(',').map { |month_stat| month_stat.split(' ').last.to_f }
```

mlshvch/hometask-3/rainfall.rb
StalemateInc: Also, this could be wrapped in a function, there is no need for it to be called in the void.

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,52 @@
...
+data = file.read
+file.close
+
+EXIT_COMMAND = 'exit!'
+ERROR_MESSAGE = 'City name can not be blank!'
+DEFAULT_VALUE = -1.0
+
+def extract_data(data_string)
+  data_string.split("\n").each_with_object({}) do |data, city_rainfall_stats|
+    town, rainfall_stats = data.split(':')
```
StalemateInc, line 15: Inconsistent naming - `town` vs `city`. Please choose one :)

mlshvch/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,52 @@
...
+
+file = File.open('data.txt', 'r')
+data = file.read
+file.close
+
+EXIT_COMMAND = 'exit!'
+ERROR_MESSAGE = 'City name can not be blank!'
+DEFAULT_VALUE = -1.0
+
+def extract_data(data_string)
```
StalemateInc, line 13: In addition to comment about `@rainfall_data` (see below): you've got options how to do it properly:
* make a function, where you read the file line by line and process it as it gets read;
* read the whole the whole at once and then use this method to extract the contents.

I like the first option more, but the second will do too.

StalemateInc (commented):
This version is clearly better, but still has some flaws to fix. I'm going to merge it, since the deadline was not met and the task (kind of) works, but I'll still ask you to fix the comments afterwards if you'll find the time, so we could get over the common misconceptions.

StalemateInc (approved):

PR #84, opened 2021-11-15
Hometask-3

Romuashi:
I added two files to this work
1) file with data about cities
2) algorithm for calculating rainfall


Romuashi/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,48 @@
...
+
+def variance(town, data)
+  return -1 unless data.include?(town)
+
+  search_town(town, data).map { |city| (mean(town, data) - city)**2 }.sum / 12
+end
+
+def search_town(town, data)
+  data.split("\n").each do |line|
+    return line.scan(/\d*\.\d/).map(&:to_f) if line.include?(town)
```
bedyan: This search doesn't work as expected, please adjust it so it'll find only if I provide correct city name

bedyan: Screenshot from 2021-11-15 21-46-56

Romuashi/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+end
+
+def run_cli
+  loop do
+    p 'Enter your string: '
```
bedyan: it's a good practice to create constants and use them instead of string or integer literals. Apply it everywhere in the code where it possible.
example:
```
ENTER_STRING = 'Enter your string: '
p ENTER_STRING

bedyan (changes requested):

bedyan (changes requested):
Screenshot from 2021-11-16 10-43-57
something wrong with the formulas(or it can be related to wrong search)
also sometimes it returns 0 if city wasn't found, not -1

bedyan:
Task wasn't done in time

bedyan (approved):

PR #86, opened 2021-11-15
Matveibos/hometask-3

Matveibos:
Домашняя работа 3


Matveibos/hometask-3/cli.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+  end
+
+  private
+
+  def data
+    @data ||= File.open('data.txt').read
+  end
+
+  def show_result(input)
```
vol6tmetr: I think that this method does two things:
1. Calculates data
2. Displays data
I think that you should separate it into two methods, with which of them having single responsibility

Matveibos/hometask-3/rainfall_statistics.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+# this class is used for calculate mean and variance values for town rainfall
+class RainfallStatistics
+  attr_reader :town, :data
```
vol6tmetr: You've defined reader methods for town and data, but didn't use it anywhere. Basically `attr_reader :town` is:
```
def town
  @town
end
```
So let's use this method instead of `@town`
Also if `town` method is not used outside of this class it should be under the private namespace

Matveibos/hometask-3/rainfall_statistics.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def initialize(town, data)
+    @town = town
+    @data = data
+    @hash = {}
```
vol6tmetr: Too bad name for a variable. From the first glance it is absolutely not clear for what this variable is used. Rename it to be more obvious, please

Matveibos/hometask-3/rainfall_statistics.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def mean
+    return -1 unless create_hash
```
vol6tmetr: Are you sure that this condition will be executed in any case?

Matveibos/hometask-3/rainfall_statistics.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    calc_mean
+  end
+
+  def variance
+    return -1 unless create_hash
```
vol6tmetr: Are you sure that this condition will be executed in any case?

Matveibos/hometask-3/rainfall_statistics.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    calc_variance
+  end
+
+  private
+
+  def create_hash
+    @data.each_line do |str|
+      key = str.split(':').first
```
vol6tmetr: Name this variable to be more obvious for what it stands for
Also move string which you use to split strings (`':'`) to the constant

Matveibos/hometask-3/rainfall_statistics.rb
```diff
@@ -0,0 +1,47 @@
...
+      array = str.scan(/\d*\.\d/).map(&:to_f)
```
vol6tmetr: Again, `array` variable does not make clear which data it contains, change it, please
Regexps are also should be moved to the constants

Matveibos/hometask-3/rainfall_statistics.rb
```diff
@@ -0,0 +1,47 @@
...
+      @hash[key] = array
+    end
+  end
+
+  def calc_mean
+    mean = @hash[@town].sum / @hash[@town].size
+    @means = mean
```
vol6tmetr: What are you doing there, exactly?

Matveibos/hometask-3/rainfall_statistics.rb
```diff
@@ -0,0 +1,47 @@
...
+  def create_hash
+    @data.each_line do |str|
+      key = str.split(':').first
+      array = str.scan(/\d*\.\d/).map(&:to_f)
+      @hash[key] = array
+    end
+  end
+
+  def calc_mean
+    mean = @hash[@town].sum / @hash[@town].size
```
vol6tmetr: You are repeating `@hash[@town]` 4 times in this class, follow DRY principle, please

Matveibos/hometask-3/rainfall_statistics.rb
```diff
@@ -0,0 +1,47 @@
...
+    @means = mean
+  end
+
+  def calc_variance
+    variance = 0
+    @hash[@town].each do |el|
```
vol6tmetr: Maybe you can rewrite this thing using `inject` method?

Matveibos/hometask-3/rainfall_statistics.rb
Matveibos: You mean hash or data?

Matveibos/hometask-3/cli.rb
vol6tmetr: You moved display logic into a different method, but you need to make this methods independent of each other. One of methods should just calculate and return calculated data, other method should display data which you pass to it

Matveibos/hometask-3/rainfall_stat.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+# this class is used for calculate mean and variance values for town rainfall
+class RainfallStat
+  SEPARATOR = ':'
+  REGULAR = /\d*\.\d/.freeze
```
vol6tmetr: Call it more explicitly, please

Matveibos/hometask-3/rainfall_stat.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def mean
+    create_hash
```
vol6tmetr: You need to have `@processed_data` variable populated for `#mean` and `#variance` methods. Why not to call this method in the initializer, rather than calling it two times in methods?

Matveibos/hometask-3/rainfall_stat.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def variance
+    create_hash
+
+    calc_variance
+  end
+
+  private
+
+  def initialize(town, data)
```
vol6tmetr: You don't need to move initializer into the private namespace. I left you a comment about reader methods, which you should move to the private namespace if they are not used outside of the class. So, let's move initializer back to the public namespace and create reader methods for values you need to access inside of this class

Matveibos/hometask-3/rainfall_stat.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def calc_mean
+    replace(@town).sum / replace(@town).size
+  end
+
+  def calc_variance
+    replace(@town).inject(0) { |var, n| var + (n - calc_mean)**2 / replace(@town).size }
+  end
+
+  def replace(town)
```
vol6tmetr: Bad name for the method, I read `#replace` method as if it is replacing something, but it is not. Better call it like `rainfall_data_for(town)`, or something like that

Matveibos/hometask-3/cli.rb
```diff
@@ -0,0 +1,36 @@
...
+# this class contains describes cli for rainfall stats
+class Cli
+  def run_cli
+    loop do
+      print ENTER_MESSAGE
+      input = gets.chomp
+
+      break if input == EXIT_MESSAGE
+
+      input.empty? ? (puts ERROR_MESSAGE) : rainfall_statistics(input)
```
vol6tmetr: Lets pass `rainfall_statistics` result straight to the `show` method
```suggestion
      input.empty? ? (puts ERROR_MESSAGE) : show(rainfall_statistics(input))
```

Matveibos/hometask-3/cli.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      show(@rainfall_statistics)
```
vol6tmetr:
```suggestion
```

Matveibos/hometask-3/rainfall_stat.rb
```diff
@@ -0,0 +1,44 @@
...
+  def variance
+    calc_variance
+  end
+
+  private
+
+  def create_hash
+    @data.each_line do |str|
+      key = str.split(SEPARATOR).first
+      array = str.scan(CHOOSING_NUMBERS).map(&:to_f)
```
vol6tmetr: I remember leaving comment to rename `key` and `array` to something more meaningful

Matveibos/hometask-3/rainfall_stat.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def initialize(town, data)
+    @town = town
+    @data = data
+    @processed_data = {}
+    create_hash
+  end
+
+  def mean
+    calc_mean
```
vol6tmetr: Do we need to have this method, just to call another method? Maybe we can just move `calc_mean` method content here?

Matveibos/hometask-3/rainfall_stat.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def variance
+    calc_variance
```
vol6tmetr: Do we need to have this method, just to call another method? Maybe we can just move `calc_variance` method content here?

Matveibos/hometask-3/rainfall_stat.rb
```diff
@@ -0,0 +1,44 @@
...
+
+# this class is used for calculate mean and variance values for town rainfall
+class RainfallStat
+  SEPARATOR = ':'
+  CHOOSING_NUMBERS = /\d*\.\d/.freeze
+
+  def initialize(town, data)
+    @town = town
+    @data = data
+    @processed_data = {}
```
vol6tmetr: Maybe we can modify `create_hash` method (btw change its name to something like `process_data`) to return processed data that we will store in the `@processed_data` variable? Wdyt?

vol6tmetr (approved):

PR #89, opened 2021-11-15
Alexey kren/hometask 3

Alexey-Kren:


ramantechart (changes requested):
This PR contains hometask-1 and hometask-2 but it has to contain only hometask-3

pavelbonda (approved):

PR #92, opened 2021-11-16
alhimick-prog/hometask-3

alhimick-prog:
Hometask-3 that requires create program, which finds the desired string, selects numerical values from it and calculates the average value and valuates of values.

This version contains three files: two files with input data and the file of the program itself. All the necessary actions for search and mathematical calculations have been implemented.

Program is started and used in accordance with technical task.


alhimick-prog/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,68 @@
...
+    file = File.new(filename, 'r')
+    file.read(file.size)
+  else
+    puts ERROR_FILE_READ_FAIL
+    exit
+  end
+end
+
+def form_array(town, strng)
+  array = nil
```
lincAlinc: @alhimick-prog `array = []`

alhimick-prog/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,68 @@
...
+  strng.each_line do |line|
+    array = line.scan(/\d+\.*\d*/).map(&:to_f) if line.match(/^#{town}:/)
+  end
+  array
+end
+
+def mean(town, strng)
+  array = form_array(town, strng)
+  return TOWN_SEARCH_ERROR_NUMBER unless array
```
lincAlinc, line 36: @alhimick-prog `array.empty?`

alhimick-prog/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  puts MESSAGE_INCORRECT_MONTH_NUMBER if array.size != MONTH_NUMBER
+  array.sum / array.size
+end
+
+def variance(town, strng)
+  array = form_array(town, strng)
+  return TOWN_SEARCH_ERROR_NUMBER unless array
```
lincAlinc, line 45: @alhimick-prog `array.empty?`

alhimick-prog/hometask-3/rainfall.rb
alhimick-prog: @lincAlinc what if the string with `town` exists but there is not a single number in the string? How do we distinguish it from a negative search result?

PR #102, opened 2021-11-18
PandaGrom hometask 4

PandaGrom:
Benchmark results

Снимок экрана от 2021-11-18 13-52-37

Benchmark  results for MyHash
Снимок экрана от 2021-11-21 21-25-11


nieisnuje (commented):
would be nice if you attach benchmark results to the pr (you can put them into the description)

PandaGrom/hometask-4/my_array_rotate_benchmark.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'benchmark'
+require './my_array_rotate'
+
+Benchmark.bm do |x|
+  puts '100'
```
nieisnuje, line 7: let's also add a benchmark for 1_000_000 elements

PandaGrom/hometask-4/my_hash_benchmark.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'benchmark'
+require './my_hash'
+
+Benchmark.bm do |x|
+  puts 'MyHash'
+  my_hash = MyHash.new
+  x.report('1_000       ') { 1_000.times { |value| my_hash[value] = value } }
```
nieisnuje, line 9: would be nice to see benchmark results in pr

nieisnuje (commented):
The code looks fine but the algorithm is not absolutely correct. 
1. You do not handle collisions well;
2. You resize the array in a strange way; `@max_density` is an unnecessary variable;

I think you can merge this task but the mark will be low.
You should study hashes better and then revisit this task.

nieisnuje (approved):

PR #109, opened 2021-11-18
Gek62/hometask-4

GEK62:


Gek62/hometask-4/my_array_rotate.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# Array class extension created
+class Array
+  def lab_rotate(displacement = 1)
+    displacement = displacement % size
+    # rubocop:disable Style/SlicingWithRange
+    self[displacement..-1] + self[0...displacement]
+    # rubocop:disable Style/SlicingWithRange
```
hrabiel: First of all, you probably wanted to reenable the previously disabled rule here, but instead, you're trying to disable it one more time here. That's why Hound is barking here.

But instead of disabling the rule, please update the code according to the Hound's previous complaints.

hrabiel (changes requested):

Gek62/hometask-4/my_array_rotate.rb
hrabiel: One more thing, please use more descriptive commit messages. Now you have 3 commits with the same message and the message itself is a bit generic which is not good

GEK62:
image
image

Gek62/hometask-4/my_array_rotate_benchmark.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+require 'benchmark'
+require './my_array_rotate'
```
hrabiel: If I run this file from a different directory, I get the error:
```
cannot load such file -- ./my_array_rotate (LoadError)
```

Using `require_relateve` should fix it

Gek62/hometask-4/my_hash_benchmark.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+require 'benchmark'
+require './my_hash'
```
hrabiel: Please, use `require_relative` here as well

Gek62/hometask-4/my_array_rotate_benchmark.rb
```diff
@@ -0,0 +1,28 @@
...
+
+Benchmark.bm do |x|
```
hrabiel, line 10: Please, make sure that you test all significant scenarios, like
- rotate by some small number (you've already done it)
- rotate by a number that is equal to array size
- rotate by a number that is more than the array's size
- rotate by a negative number

Gek62/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+# Hash class extension created
+class MyHash
```
hrabiel, line 4: Please, rewrite and make sure that your hash works exactly the same way as `Hash`, because at the moment it does not.

hrabiel (changes requested):

Gek62/hometask-4/my_hash_benchmark.rb
```diff
@@ -0,0 +1,26 @@
...
+  x.report('read(100)') { @my_hash[100] }
+  x.report('delete') { @my_hash.delete(100) }
+  x.report('length') { @my_hash.length }
+  puts 'Hash'
+  x.report('10') { 10.times { |i| @hash[i] = i } }
+  x.report('100') { 100.times { |i| @hash[i] = i } }
+  x.report('1000') { 1000.times { |i| @hash[i] = i } }
+  x.report('10000') { 10_000.times { |i| @hash[i] = i } }
+  x.report('read(100)') { hash[100] }
+  x.report('delete') { @hash.clear }
```
hrabiel: When benchmarking `MyHash`, you tested "delete item by key" method, but here you're testing "delete all items" method. It looks inconsistent.
You should compare `MyHash#delete` with `Hash#delete` and `MyHash#delete_all` with `Hash#clear`.

Gek62/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,92 @@
...
+    collision = slot(key)
+
+    hash_pair(collision, key)&.last
+  end
+
+  def delete(key)
+    @pair_count -= 1
+
+    deleted_pair = slot(key).yield_self do |collision|
+      collision.dup - collision.delete_if { |k, _| k == key }
```
hrabiel: I see 2 problems here. They're not critical at all, but anyway.
First of all, you don't have to use a placeholder argument (`_`) here, it's just redundant.
And moreover, `delete_if` method expects a block accepting _just one_ argument, so I can say it's redundant twice. :slightly_smiling_face:

Gek62/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,92 @@
...
+      collision.each { |k, v| self[k] = v }
+    end
+  end
+
+  def hash_key(key)
+    key.hash % size
+  end
+
+  def hash_pair(slot, key)
+    slot.find { |k, _| k == key }
```
hrabiel: The same issue as with the usage of `delete_if` method above.

hrabiel (changes requested):

hrabiel (approved):

PR #113, opened 2021-11-18
Mlshvch hometask 4

mlshvch:
I've completed hometask 4 with described requirements


mlshvch/hometask-4/my_array_rotate.rb
```diff
@@ -0,0 +1,31 @@
...
+# class Array is extended with methods lab_rotate and lab_rotate! methods
+# with the Array.rotate and Array.rotate! functionality respectively
+class Array
+  def adjust_array_shift(shift, array_length)
+    if shift.positive?
+      shift % array_length
+    else
+      -(-shift % array_length)
+    end
+  end
```
StalemateInc: First, this should be a private method. Second, I think it can fit in a ternary form.

mlshvch/hometask-4/my_array_rotate.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def lab_rotate(shift = 1)
+    return self if shift.zero?
+
+    shift = adjust_array_shift(shift, length)
+    if shift.positive?
+      # method Array#append is not used because of Rubocop Assignment Condition Size Branch is too high
```
StalemateInc: Let's write such notes in PR, not in the code itself.

mlshvch/hometask-4/my_array_rotate.rb
```diff
@@ -0,0 +1,31 @@
...
+      self[shift..(length - 1)] + self[0..(shift - 1)]
```
StalemateInc, line 13: You can use a beginless range here as well.
```suggestion
      self[shift..(length - 1)] + self[..(shift - 1)]
```

mlshvch/hometask-4/my_array_rotate.rb
```diff
@@ -0,0 +1,31 @@
...
+    else
+      self[shift..] + self[0..(length + shift - 1)]
```
StalemateInc, line 15: And here
```suggestion
      self[shift..] + self[..(length + shift - 1)]
```

mlshvch/hometask-4/my_array_rotate.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+  end
+
+  def lab_rotate!(shift = 1)
+    replace(lab_rotate(shift))
```
StalemateInc, line 20: 👍

mlshvch/hometask-4/my_array_rotate_benchmark.rb
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env ruby
+
+# frozen_string_literal: true
+
+require 'benchmark'
+require './my_array_rotate'
+
+puts "\n100"
```
StalemateInc: It would be nice to understand what exactly this number means. Something like "Benchmarking with 100 elements".

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env ruby
+
+# frozen_string_literal: true
+
+require 'digest'
+# class MyHash with custom implementation of Hash Table
```
StalemateInc, line 5: Insert a newline before requires and the beginnings of classes / modules.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+class MyHash
+  include Enumerable
+  EMPTY_HASH_TABLE_SIZE = 10
```
StalemateInc, line 9: Insert a newline between included modules and constants definitions.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  def initialize
+    create_empty_hash
+  end
+
+  def create_empty_hash(hash_table_size = EMPTY_HASH_TABLE_SIZE)
+    @hash_table_size = hash_table_size
+    @hash_table = Array.new(EMPTY_HASH_TABLE_SIZE, [])
+    @number_of_entries = 0.0
+    self
```
StalemateInc: No need to return `self`. Instance variables are modifying the state of the object and nothing really expected to be returned out of this method. If you are worrying about `initialize`, then the return value of that method is always ignored.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+  end
+
+  def calculate_hash_table_index(key)
+    # sha = Digest::SHA2.hexdigest(key.to_s)
+    # hash = ''
+    # 10.times do |index|
+    #   hash += sha[index].ord.to_s
+    # end
+    # hash.to_i % @hash_table_size
```
StalemateInc: No useless comments, please :)

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+# class MyHash with custom implementation of Hash Table
+class MyHash
+  include Enumerable
+  EMPTY_HASH_TABLE_SIZE = 10
+
+  def initialize
+    create_empty_hash
+  end
+
+  def create_empty_hash(hash_table_size = EMPTY_HASH_TABLE_SIZE)
```
StalemateInc: Do you really need `hash_table_size` param? I think even after the whole clear of hash, hash should shrink the amount of buckets back to its default value.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  def []=(key, value)
+    index = calculate_hash_table_index(key)
+    @hash_table[index] = if @hash_table[index].to_a.empty? || @hash_table[index].to_a.assoc(key)
+                           [[key, value]]
+                         else
+                           @hash_table[index].to_a.append([key, value])
+                         end
+    @number_of_entries += 1
+    rehash_table
```
StalemateInc: Rehashing the table **after** you've inserted the value leads to at least value inserted into a wrong bucket.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+  end
+
+  def create_empty_hash(hash_table_size = EMPTY_HASH_TABLE_SIZE)
+    @hash_table_size = hash_table_size
+    @hash_table = Array.new(EMPTY_HASH_TABLE_SIZE, [])
+    @number_of_entries = 0.0
+    self
+  end
+
+  def calculate_hash_table_index(key)
```
StalemateInc: Do you need to call this method outside of the hash?

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+  end
+
+  def []=(key, value)
+    index = calculate_hash_table_index(key)
+    @hash_table[index] = if @hash_table[index].to_a.empty? || @hash_table[index].to_a.assoc(key)
+                           [[key, value]]
+                         else
+                           @hash_table[index].to_a.append([key, value])
+                         end
+    @number_of_entries += 1
```
StalemateInc: So, if I add a key-value pair to hash, while the same key already exists, the amount of key-pairs will remain the same, but counter will be increased anyway? That's a bug.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+    # hash.to_i % @hash_table_size
+    key.hash % @hash_table_size
+  end
+
+  def []=(key, value)
+    index = calculate_hash_table_index(key)
+    @hash_table[index] = if @hash_table[index].to_a.empty? || @hash_table[index].to_a.assoc(key)
+                           [[key, value]]
+                         else
+                           @hash_table[index].to_a.append([key, value])
```
StalemateInc: Why do you need to cast arrays to arrays? (I mean `.to_a` calls)

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+    # 10.times do |index|
+    #   hash += sha[index].ord.to_s
+    # end
+    # hash.to_i % @hash_table_size
+    key.hash % @hash_table_size
+  end
+
+  def []=(key, value)
+    index = calculate_hash_table_index(key)
+    @hash_table[index] = if @hash_table[index].to_a.empty? || @hash_table[index].to_a.assoc(key)
```
StalemateInc: > if @hash_table[index].to_a.empty? || @hash_table[index].to_a.assoc(key)

Why do you need to check for emptiness first? You will get the same outcome with find / assoc anyway.

> assoc(key)

Nice, I did not remember about such method myself.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+                           [[key, value]]
```
StalemateInc: When inserting values with the same key, new value should replace the old one and not be inserted with the duplicate key once again into the same bucket.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+                         else
+                           @hash_table[index].to_a.append([key, value])
+                         end
+    @number_of_entries += 1
+    rehash_table
+  end
+
+  def [](key)
+    index = calculate_hash_table_index(key)
+    @hash_table[index].to_a.assoc(key).last unless @hash_table[index].to_a.empty?
```
StalemateInc: Guard clause would play nicer here. Smth like:
```ruby
pair = @hash_table[index].to_assoc(key)
return nil unless pair

pair.last
```
A bit easier to process :)

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+    nil
+  end
+
+  def clean
+    create_empty_hash
+  end
+
+  def load_factor
+    @number_of_entries / @hash_table_size
+  end
```
StalemateInc: Do you need to call this method outside of `MyHash` internals? It has to be a private method.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  def show
+    @hash_table.each_index do |index|
+      puts "#{index}: #{@hash_table[index]}"
+    end
+  end
```
StalemateInc: This method is not required, please delete.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+    old_hash_table = Array.new(@hash_table_size, []).replace(@hash_table)
+    old_hash_table_size = @hash_table_size
+    create_empty_hash(old_hash_table_size * 2)
+    old_hash_table.each_index do |index|
+      next if old_hash_table[index].nil?
+
+      old_hash_table[index].each { |key_value_pair| restore(key_value_pair[0], key_value_pair[1]) }
+    end
+    @hash_table
+  end
```
StalemateInc: This has to be a private method.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+      puts "#{index}: #{@hash_table[index]}"
+    end
+  end
+
+  def size
+    @number_of_entries
+  end
+
+  def rehash_table
+    return unless load_factor > 0.7
```
StalemateInc: This check can be extracted in a separate method like `requires_resizing?`

StalemateInc: Also magic constants, again.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+    rehash_table
+  end
+
+  def [](key)
+    index = calculate_hash_table_index(key)
+    @hash_table[index].to_a.assoc(key).last unless @hash_table[index].to_a.empty?
+  end
+
+  def delete(key)
+    @number_of_entries -= 1
```
StalemateInc: Why do you lower the amount of entries before actually deleting a value? What if an error occurs after this line? The element won't be deleted and your counter will remain incorrect.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+    index = calculate_hash_table_index(key)
+    @hash_table[index].to_a.each_with_index do |elem, bucket_index|
+      return @hash_table[index].delete_at(bucket_index).last if elem[0] == key
+    end
+    @number_of_entries += 1
+    nil
```
StalemateInc: Since we mimic `Hash` behaviour, let's return deleted value from this method.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+    @hash_table[index].to_a.assoc(key).last unless @hash_table[index].to_a.empty?
+  end
+
+  def delete(key)
+    @number_of_entries -= 1
+    index = calculate_hash_table_index(key)
+    @hash_table[index].to_a.each_with_index do |elem, bucket_index|
+      return @hash_table[index].delete_at(bucket_index).last if elem[0] == key
+    end
+    @number_of_entries += 1
```
StalemateInc: I don't understand why do we increase the number of elements in a `delete` method.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+    index = calculate_hash_table_index(key)
+    @hash_table[index].to_a.assoc(key).last unless @hash_table[index].to_a.empty?
+  end
+
+  def delete(key)
+    @number_of_entries -= 1
+    index = calculate_hash_table_index(key)
+    @hash_table[index].to_a.each_with_index do |elem, bucket_index|
+      return @hash_table[index].delete_at(bucket_index).last if elem[0] == key
+    end
```
StalemateInc: This is definitely hard to read. Overengineering is equally as bad as not thinking about the solution enough. Just find the index of an item in a bucket and delete it if it exists.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+  EMPTY_HASH_TABLE_SIZE = 10
+
+  def initialize
+    create_empty_hash
+  end
+
+  def create_empty_hash(hash_table_size = EMPTY_HASH_TABLE_SIZE)
+    @hash_table_size = hash_table_size
+    @hash_table = Array.new(EMPTY_HASH_TABLE_SIZE, [])
+    @number_of_entries = 0.0
```
StalemateInc: Can you get a fraction of an entry in a hash? Like 1.2 entries?

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+    @number_of_entries += 1
+    nil
+  end
+
+  def clean
+    create_empty_hash
+  end
+
+  def load_factor
+    @number_of_entries / @hash_table_size
```
StalemateInc: Load factor has to be a float value. Both of used values should be integers, so regular `/` method won't work correctly.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+
+    old_hash_table = Array.new(@hash_table_size, []).replace(@hash_table)
+    old_hash_table_size = @hash_table_size
+    create_empty_hash(old_hash_table_size * 2)
+    old_hash_table.each_index do |index|
+      next if old_hash_table[index].nil?
+
+      old_hash_table[index].each { |key_value_pair| restore(key_value_pair[0], key_value_pair[1]) }
+    end
+    @hash_table
```
StalemateInc: a) Why can't you use the `@hash_table` directly in line 79?
b) You don't have to return anything from this method, you are using it to change the internal state of your hash
c) Ideally, you have to just save locally your buckets, expand you current hash and go through each old bucket copying the key-value pairs, preliminarily rehashing the values.

It's too convoluted and makes lots of things needlessly complex. Let's think of improving this.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,101 @@
...
+  def restore(key, value)
+    index = calculate_hash_table_index(key)
+    @hash_table[index] = if @hash_table[index].to_a.empty? || @hash_table[index].to_a.assoc(key)
+                           [[key, value]]
+                         else
+                           @hash_table[index].to_a.append([key, value])
+                         end
+    @number_of_entries += 1
+  end
+end
```
StalemateInc: This is the exact copy of `#[]=` method? There is no need to duplicate the same code across the class. Just reuse what you can reuse.

StalemateInc (changes requested):
Where is the README?

mlshvch/hometask-4/my_hash.rb
mlshvch: Nice idea)

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env ruby
+
+# frozen_string_literal: true
+
+require 'digest'
```
StalemateInc: No need to require `digest` if you don't use it.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,89 @@
...
+      @number_of_entries += 1
+    elsif @hash_table[index].assoc(key)
+      @hash_table[index].each { |bucket_elem| bucket_elem[1] = value if bucket_elem[0] == key }
+    else
+      @hash_table[index] << [key, value]
+      @number_of_entries += 1
+    end
+  end
+
+  def create_empty_hash(hash_table_size = EMPTY_HASH_TABLE_SIZE)
```
StalemateInc: You never pass anything to this param, please consider removing it completely. YAGNI principle in action.

StalemateInc: I would also recommend to make move the body to `#clean` method.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  def initialize
+    create_empty_hash
+  end
+
+  def []=(key, value)
+    rehash_table if requires_resizing?
+    index = calculate_hash_table_index(key)
+    write_to_hash(index, key, value)
+    @hash_table[index].assoc(key)[1]
```
StalemateInc: You don't really need to perform a lookup after writing to a hash, makes you writes at least N operations slower. All it takes is returning the inserted value from `write_to_hash` method.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,89 @@
...
+  end
+
+  def size
+    @number_of_entries
+  end
+
+  private
+
+  def requires_resizing?
+    (@number_of_entries.to_f / @hash_table_size) > HASH_LOAD_FACTOR
```
StalemateInc: This way it also works, but you can also take a look at [#fdiv method](https://devdocs.io/ruby~3/integer#method-i-fdiv).

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,89 @@
...
+  end
+
+  def rehash_table
+    @hash_table += Array.new(@hash_table_size, [])
+    @hash_table_size *= 2
+    @number_of_entries = 0
+    @hash_table.each_index do |index|
```
StalemateInc: Why not plain `each`?
```suggestion
    @hash_table.each do |bucket|
```
and use bucket instead of accessing @hash_table[index] again and again.

The biggest concern I wanted to raise is the amount of operations performed.
Your solution modifies the current hash table and then goes through the values, moving them to new places by removing from original bucket and adding to the new one. I think it could be slightly better if you created the new hash table first, copied the values over to it and then just swapped one @hash_table with another.

Let's see an example:
Let's pretend that we have a hash of 100 elements, each in their own bucket. In worst case your solution it will take around 200 * (1 deletion + 1 store) = 400 operations. (we will pretend that resizing the array and hashing is free at the moment and worst case means that rehashed values will shift by 100 elements, i.e. 1 element will become 101).
While the "recreate and copy" example will take only a 100 operations, since you iterate only through existing values and just put rehashed value into appropriate bucket.

Does it sound better, what do you think?

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,89 @@
...
+  def write_to_hash(index, key, value)
+    if @hash_table[index].empty?
+      @hash_table[index] = [[key, value]]
+      @number_of_entries += 1
+    elsif @hash_table[index].assoc(key)
+      @hash_table[index].each { |bucket_elem| bucket_elem[1] = value if bucket_elem[0] == key }
+    else
+      @hash_table[index] << [key, value]
+      @number_of_entries += 1
+    end
```
StalemateInc: Your top and bottom conditions are basically equal, one of them has to go away.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,89 @@
...
+      end
+    end
+  end
+
+  def calculate_hash_table_index(key)
+    key.hash % @hash_table_size
+  end
+
+  def write_to_hash(index, key, value)
+    if @hash_table[index].empty?
```
StalemateInc: You need `@hash_table[index]` in all three cases. How about seeking for it first, saving into a variable and then using into next calculations?

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,89 @@
...
+      @hash_table[index] = [[key, value]]
+      @number_of_entries += 1
+    elsif @hash_table[index].assoc(key)
+      @hash_table[index].each { |bucket_elem| bucket_elem[1] = value if bucket_elem[0] == key }
```
StalemateInc: All keys in each buckets should be unique by design, there is no need to iterate through all of them with `each`, just find the one you need.

StalemateInc (commented):
Close enough, but storing the value and rehashing the contents are still not quite optimal, please see the comments below:

mlshvch/hometask-4/my_hash.rb
mlshvch: So I did a commit ago! Maybe I haven't understand you properly (paragraph C)

StalemateInc: Don't get frustrated, it happens sometimes. The previous mistake was to merge the "old" hash table into "new" one just to immediately rehash the values and move them somewhere else, while I suggested creating an expanded version of the hash table, copying the from old to new values with rehashing and then swapping the hash tables.
It's actually very easy.

mlshvch/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env ruby
+
+# frozen_string_literal: true
+
+require 'digest'
+
+# class MyHash with custom implementation of Hash Table
+class MyHash
+  include Enumerable
```
StalemateInc, line 7: By the way, I see you included `Enumerable`, but where is the `each` method?

mlshvch/hometask-4/my_hash.rb
mlshvch: I don't understand: do I need to copy hash table to variable, clear the hash, and after that insert every key-value pair pair from variable back to hash?

mlshvch/hometask-4/my_hash.rb
mlshvch: I also want to know. Other methods such as each_with_object are available according to MyHash#methods

mlshvch/hometask-4/my_hash.rb
mlshvch: This is a necessity: otherwise error  
`не получается пофиксить баг:
@hash_table[index] << [key, value]
записывает пару в КАЖДЫЙ бакет`
will occur

mlshvch/hometask-4/my_hash.rb
mlshvch: This branch is for changing value by the key. I have to find the one with each

mlshvch/hometask-4/my_hash.rb
StalemateInc: > не получается пофиксить баг: @hash_table[index] << [key, value] записывает пару в КАЖДЫЙ бакет

Well, I'm not really supposed to fix bugs for you, but I think I can help with this one. Take a look [at this line](https://github.com/iTechArt-RubyLab/Ruby-Lab/blob/1301955ed4b8008ebf251399394dca53dbebfb1a/mlshvch/hometask-4/my_hash.rb#L86) and then at the ["Creating arrays" section](https://devdocs.io/ruby~3/array). You will see that the second argument for array constructor is a default object, an object, which will be inserted instead of nils. You are not inserting the same value into each bucket, as you may think, you actually insert the value once, but each of your buckets is the same object. If you want to initialize all values you need to use a block (it is described in the same section). Like this:
```ruby
 Array.new(EMPTY_HASH_TABLE_SIZE) { [] }
```

mlshvch/hometask-4/my_hash.rb
StalemateInc: Please read the Enumerable module documentation carefully:
"The class must provide a method each, which yields successive members of the collection." <- you have to implement it by yourself.

The presence of other methods does not mean that they actually work, try calling `map` on your hash and you will see smth like `NoMethodError (undefined method 'each' for #<MyHash:...>)`

mlshvch/hometask-4/my_hash.rb
StalemateInc: If by "clearing the hash" you mean "replace with expanded, but empty hash table", then yes, you are correct.

mlshvch/hometask-4/my_hash.rb
StalemateInc: > I have to find the one with each

But you've found it already - it is `@hash_table[index].assoc(key)` if it's not `nil`.

mlshvch/hometask-4/my_hash.rb
mlshvch: Thanks)

mlshvch/hometask-4/my_hash.rb
mlshvch: Nice idea!

mlshvch/hometask-4/my_hash.rb
mlshvch: Understood

StalemateInc (approved):
Nicely done 👍

PR #114, opened 2021-11-18
TehanovAnton/hometask-4

TehanovAnton:


TehanovAnton/hometask-4/my_hash_benchmark.rb
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env ruby
```
dmokreckiy: Why shebang line?

TehanovAnton/hometask-4/my_hash_benchmark.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+end
+
+Benchmark.bm do |x|
+  hash_benchmark = HashBenchmark.new({}, x)
+  my_hash_benchmark = HashBenchmark.new(MyHash.new, x)
+
+  hash_benchmark.write_report(100)
+  my_hash_benchmark.write_report(100)
+  puts ''
```
dmokreckiy: Question - do we need empty string here?

TehanovAnton/hometask-4/my_array_rotate_benchmark.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
```
dmokreckiy, line 1: No shebang here. Generalize it across both benchmarks.

TehanovAnton/hometask-4/my_hash.rb
```diff
@@ -0,0 +1,90 @@
...
+    pair = pair(key)
+
+    pair&.last
+  end
+
+  def []=(key, value)
+    resize! if resize?
+
+    slot = slot(key)
+    pair = pair(key, slot: slot)
```
dmokreckiy: Do you think it is a good idea to name local variables with same names as private methods?

dmokreckiy (dismissed):
LGTM
Check out my comments.
Also we need to discuss implementation.

dmokreckiy (approved):
LGTM

PR #115, opened 2021-11-20
Valeriyapetrova/hometask 5

ghost:
Hometask-5

### DIFF
--- a/valeriyapetrova/hometask-5/4.1.sql
+++ b/valeriyapetrova/hometask-5/4.1.sql
@@ -1,0 +1,1 @@
+SELECT * FROM actor WHERE last_name = 'Zellweger';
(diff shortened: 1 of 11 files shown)


valeriyapetrova/hometask-5/4.1.sql
```diff
@@ -0,0 +1 @@
+SELECT * FROM actor WHERE last_name IN ('Zellweger');
```
TanyaAkulich: `... last_name IN ('Zellweger')` is fine but it also may be replaced with `last_name = 'Zellweger'`

valeriyapetrova/hometask-5/4.8.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM film WHERE rental_rate = 0.99;
```
TanyaAkulich: cascade

TanyaAkulich (changes requested):
you've missed cascade delete (it's must)

empty line in the end isn't required but git's alerts are too annoying, please try to not ignore it in the future

and plase order the commit history

TanyaAkulich (approved):

PR #117, opened 2021-11-21
alhimick-prog/hometask-5

alhimick-prog:
Hometask 5: working with databases.
Files with SQL query.


alhimick-prog/hometask-5/4.2.sql
```diff
@@ -0,0 +1 @@
+SELECT title FROM film WHERE language_id IN (1, 6, 2);
```
lincAlinc: @alhimick-prog `IN (1, 6, 2)` looks like chittering 😄 What if I will change ids for these languages?

lincAlinc (changes requested):

alhimick-prog/hometask-5/4.2.sql
alhimick-prog: @lincAlinc i may be a cheater, but I'm not chittering.

lincAlinc (approved):
Looks good

PR #120, opened 2021-11-21
Upload hometask-3

AlexeyShparun:
Upload hometask-3

### DIFF
--- a/AlexeyShparun/hometask-3/rainfall.rb
+++ b/AlexeyShparun/hometask-3/rainfall.rb
@@ -8,22 +8,22 @@ end
 
 def string_search(town, string)
   city = string.split("\n").find { |current_string| current_string.include? town }
-  return nil if city.nil?
+  return if city.nil?
 
   city.scan(/\d*\.\d/).map(&:to_f)
 end
 
 def mean(town, string)
-  return -1 if (average = string_search(town, string)).nil?
+  @average = string_search(town, string)
+  return -1 if @average.nil?
 
-  average.sum / 12
+  @mean = @average.sum / 12
 end
 
-def variance(town, string)
-  return -1 if (average = string_search(town, string)).nil?
+def variance
+  return -1 if @average.nil?
 
-  mean = average.sum / 12
-  average.map { |value| (mean - value)**2 }.sum / 12
+  @average.map { |value| (@mean - value)**2 }.sum / 12
 end
 
 def run_cli
@@ -42,7 +42,7 @@ end
 
 def result(input)
   data = upload_data
-  puts "Mean: #{mean(input, data)}\nVariance: #{variance(input, data)}"
+  puts "Mean: #{mean(input, data)}\nVariance: #{variance}"
 end
 
 run_cli
--- a/AlexeyShparun/hometask-3/rainfall.rb
+++ b/AlexeyShparun/hometask-3/rainfall.rb
@@ -1,37 +1,29 @@
 #!/usr/bin/env ruby
+
 # frozen_string_literal: true
 
 def upload_data
-  data = File.new('data.txt')
-  content = data.read
-  data.close
-  content
+  File.read('data.txt')
 end
 
-def string_search(town, _string)
-  val = upload_data.split("\n").find { |town_val| town_val.include? town }
-  x = ''
-  val.nil? ? x : val
-end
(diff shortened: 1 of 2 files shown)


AlexeyShparun/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+def uploading_data
```
fayriss: it's better to have namings with infinitive verbs, i.e. `upload_data`

AlexeyShparun/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,46 @@
...
+  data = File.new('data.txt')
+  data.read
+end
+
+def mean(town, string)
+  return -1 unless string.include?(town)
+
+  city_search(town, string).sum / 12
```
fayriss: you are doing heavy operations like `city_search` and `uploading_data` for the same params every time

consider saving results in variables

AlexeyShparun/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+def uploading_data
+  data = File.new('data.txt')
+  data.read
+end
+
+def mean(town, string)
+  return -1 unless string.include?(town)
```
fayriss: 1) what does `return -1` mean for error handling? 
2)it's better to move city validation to corresponding `city_search` method

same comments apply to this line in `variance` method

AlexeyShparun/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+def uploading_data
+  data = File.new('data.txt')
+  data.read
```
fayriss: you need to close the file after reading

AlexeyShparun/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,46 @@
...
+def mean(town, string)
+  return -1 unless string.include?(town)
+
+  city_search(town, string).sum / 12
+end
+
+def variance(town, string)
+  return -1 unless string.include?(town)
+
+  city_search(town, string).map { |x| (mean(town, string) - x)**2 }.sum / 12
```
fayriss: since you use results from previous action, consider saving results somewhere

AlexeyShparun/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,46 @@
...
+end
+
+def city_search(town, string)
+  string.split("\n").each do |city|
+    return city.scan(/\d*\.\d/).map(&:to_f) if city.include?(town)
```
fayriss: `.find { ... }.then { ... }` syntax might look cleaner in this case

fayriss (changes requested):

AlexeyShparun/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,48 @@
...
+
+def string_search(town, string)
+  city = string.split("\n").find { |current_string| current_string.include? town }
+  return nil if city.nil?
+
+  city.scan(/\d*\.\d/).map(&:to_f)
+end
+
+def mean(town, string)
+  return -1 if (average = string_search(town, string)).nil?
```
fayriss: You should call `string_search` just once, the same as `upload_data`

fayriss: it's better to have assignment in a separate line

```ruby
average = string_search(town, string)
return -1 if average.nil?

```

AlexeyShparun/hometask-3/rainfall.rb
```diff
@@ -0,0 +1,48 @@
...
+
+# frozen_string_literal: true
+
+def upload_data
+  File.read('data.txt')
+end
+
+def string_search(town, string)
+  city = string.split("\n").find { |current_string| current_string.include? town }
+  return nil if city.nil?
```
fayriss: you can just do `return if city.nil?`

fayriss (approved):

PR #121, opened 2021-11-21
PandaGrom hometask 5

PandaGrom:


nieisnuje (changes requested):
why do you have commits from task 4 here?
I think you should push task 5 into a separate branch which is created from the main repository's master, not from your branch for task 4

PandaGrom/hometask-5/4.3.sql
```diff
@@ -0,0 +1 @@
+SELECT CONCAT(last_name, '-', first_name) FROM actor WHERE first_name like '%W' OR last_name like 'H%'
```
nieisnuje, line 1: are you sure that here you select names starting with W and surnames ending with H?

PandaGrom/hometask-5/4.5.sql
```diff
@@ -0,0 +1,2 @@
+SELECT name, COUNT(language.language_id) FROM film
```
nieisnuje, line 1: when you write a query with joins it's better to specify attributes of which tables you are selecting
like `language.name` not just `name`

PandaGrom/hometask-5/4.7.sql
```diff
@@ -0,0 +1,2 @@
+ UPDATE film SET release_year = 2020, rental_rate = 9  WHERE title = 'The Gentlemen'
+
```
nieisnuje, line 2: remove empty lines

PandaGrom/hometask-5/4.8.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM film cascade WHERE title = 'The Gentlemen'
```
nieisnuje, line 1: I don't think you should specify `cascade` option while deleting
Please, doublecheck in the documentation

PR #122, opened 2021-11-22
Gek62/hometask-5

GEK62:


Gek62/hometask-5/5.4.2.sql
```diff
@@ -0,0 +1,4 @@
+SELECT *
+FROM film
+WHERE language_id IN (1,2,6)
```
hrabiel: I suggest not to use hardcoded language_id's here.
The better way is to specify languages by name. Otherwise, it's not clear what the request does.
To achieve that you have to use either a join or subselect.
Please, do it.

Gek62/hometask-5/5.4.3.sql
```diff
@@ -0,0 +1,3 @@
+SELECT CONCAT(last_name, '-', first_name) AS full_name
+FROM actor
+WHERE first_name LIKE '%W' OR last_name LIKE 'H%'
```
hrabiel: According to the task requirements, the first name should **_start_** with W and the last name should **_end_** with H.
That's not what you did.

Also, take into consideration case sensitivity. In my opinion, the best approach is to make the patterns case-insensitive.

Gek62/hometask-5/5.4.8.sql
```diff
@@ -0,0 +1,3 @@
+DELETE 
+FROM film CASCADE
```
hrabiel: Does it even work?
I think, no. Try a different year, which exists in the database (2013 does not), and you'll see an error caused by the foreign key constraint.

As far as I know, If we don't define `ON DELETE CASCADE` rule in the foreign key constraint, we should manually delete all related records.

Gek62/hometask-5/5.4.9.sql
```diff
@@ -0,0 +1,6 @@
+SELECT actor.first_name, actor.last_name, COUNT(film_actor.film_id)
+FROM actor 
+INNER JOIN film_actor ON actor.actor_id = film_actor.actor_id 
+GROUP BY actor.first_name, actor.last_name
```
hrabiel: What if there are two or more actors with the same first and last name? By the way, it's exactly the case in this database.
So I think it's better to group by actor ID, do you agree?

Gek62/hometask-5/5.4.1.sql
```diff
@@ -0,0 +1,3 @@
+SELECT *
+FROM actor
+WHERE last_name = 'Zellweger';
```
hrabiel: Please add newlines at the end of each file, this is a common practice

hrabiel (changes requested):
In general, good job! :+1:

Gek62/hometask-5/5.4.3.sql
```diff
@@ -0,0 +1,3 @@
+SELECT CONCAT(last_name, '-', first_name) AS full_name
+FROM actor
+WHERE first_name LIKE 'w%' OR last_name LIKE '%h'
```
hrabiel: Unfortunately, this is still not case-insensitive, it works with lower case only.
Consider using Postgres' `ILIKE` operator or `LOWER()`.

hrabiel (changes requested):

hrabiel (approved):

PR #124, opened 2021-11-22
PandaGrom hometask 5

PandaGrom:


nieisnuje (approved):
All remarks were fixed in the previous PR, now LGTM

PR #131, opened 2021-11-22
mlshvch hometask 5

mlshvch:
I've completed hometask 5 with described requirements


mlshvch/hometask-5/1.sql
```diff
@@ -0,0 +1,3 @@
+SELECT CONCAT(first_name, ' ', last_name) AS actors
```
StalemateInc: a) The alias is not needed if you are not referencing it anywhere;
b) It's not clearly stated in the task, but I think `SELECT * ...` should be enough. Otherwise you've lost the `actor_id` and `last_updated` columns. But once again, the exact list of returned columns is not specified in the task, so it's a minor thing.

mlshvch/hometask-5/1.sql
```diff
@@ -0,0 +1,3 @@
...
+FROM actor
+WHERE last_name = 'Zellweger';
```
StalemateInc: Let's build a healthy habit of leaving an empty newline in the end of each file.

mlshvch/hometask-5/2.sql
```diff
@@ -0,0 +1,5 @@
+SELECT f.title, l.name
+FROM film f
```
StalemateInc: Isn't it much easier to just write "film" instead of random letter? Let's avoid unnecessary shortening and keep the code as much readable as possible.
Applies to the rest of files as well.

mlshvch/hometask-5/2.sql
```diff
@@ -0,0 +1,5 @@
...
+         LEFT JOIN language l
+                   ON f.language_id = l.language_id
```
StalemateInc: That's purely subjective, but I don't see the issue with JOIN clause and ON being on the same line
```suggestion
LEFT JOIN language l ON f.language_id = l.language_id
```

mlshvch/hometask-5/3.sql
```diff
@@ -0,0 +1,3 @@
+SELECT CONCAT(last_name, '-', first_name)
+FROM actor
+WHERE first_name LIKE 'W%' OR last_name LIKE '%h';
```
StalemateInc: Note: if you don't need to care about case, you can use case-insensitive `ILIKE` instead of regular `LIKE`.

mlshvch/hometask-5/8.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE film_actor
+DROP CONSTRAINT film_actor_film_id_fkey;
+
+ALTER TABLE film_category
+DROP CONSTRAINT film_category_film_id_fkey;
+
+ALTER TABLE inventory
+DROP CONSTRAINT inventory_film_id_fkey;
```
StalemateInc: Why do you need to drop constraints? This task does not involve changing DB structure.

mlshvch/hometask-5/8.sql
```diff
@@ -0,0 +1,11 @@
...
+
+delete from film
+where length > 150;
```
StalemateInc: Let's be consistent and stick one SQL notation, preferably one with keywords in all caps.

mlshvch/hometask-5/9.sql
```diff
@@ -0,0 +1,9 @@
+SELECT CONCAT(a.first_name, ' ', a.last_name) AS name, FA.count
+FROM actor AS a
+         LEFT JOIN
+     (SELECT fa.actor_id, COUNT(fa.actor_id) as count
+      FROM film_actor AS fa
+      GROUP BY fa.actor_id
+         ) AS FA
```
StalemateInc: You don't need a subquery here, you can get away with JOIN, HAVING and GROUP BY.

StalemateInc (changes requested):
Overall looks fine, but still contains a few places to fix. Please pay attention to №8 and №9 - they've been completed incorrectly.

mlshvch/hometask-5/8.sql
mlshvch: Deletion is blocked because of foreign keys

StalemateInc: Foreign keys exist to maintain database integrity and doing their job of preventing you from deleting records and leaving the dependent records pointing to the nonexistent ones. I recommend to get back to the PostgreSQL docs and pay attention to DELETE command options, the answer is there.

mlshvch: Do you mean ONLY option?

StalemateInc: No, I mean CASCADE option, but somehow I can't find it in the docs

mlshvch: Honestly, I had a thought about cascade, but wasn't sure

StalemateInc: FYI: there is no cascade option, that's right. You should probably delete other references before actually deleting the films. Foreign keys themselves should remain in place.

mlshvch: But at this moment I don't understand how to use it: this option has to be added during DB design, so on `DELETE` command everything has to happen automatically without any additional parameters

StalemateInc: Yes, please see the my previous comment. 😉

mlshvch: Sure) Today something with Internet or GitHub on my side: delays happen frequently

mlshvch/hometask-5/8.sql
```diff
@@ -0,0 +1,10 @@
+DELETE FROM inventory
+WHERE inventory.film_id IN (SELECT film.film_id FROM film WHERE length > 150 );
+
+DELETE FROM film_category
+WHERE film_id IN (SELECT film.film_id FROM film WHERE length > 150 );
+
+DELETE FROM film_actor
+WHERE film_id IN (SELECT film.film_id FROM film WHERE length > 150 );
```
StalemateInc, line 8: I am not going to torture you with this since there isn't much time left, but the better way to write this query was to use USING option:
```suggestion
DELETE FROM film_actor USING film WHERE film_actor.film_id = film.film_id AND film.length > 120;
```

StalemateInc (approved):
Good job 👍

mlshvch/hometask-5/8.sql
mlshvch: Nothing is better than pure SQL :)

PR #137, opened 2021-11-22
silitski_hometask/5

qqqwww123cdv:
Finished 5 homework.

### DIFF
--- a/qqqwww123cdv/hometask-5/3.sql
+++ b/qqqwww123cdv/hometask-5/3.sql
@@ -1,0 +1,3 @@
+SELECT last_name || '-' || first_name AS full_name
+FROM actor
+WHERE first_name LIKE 'W%' OR last_name LIKE '%H';
--- a/qqqwww123cdv/hometask-5/3.sql
+++ b/qqqwww123cdv/hometask-5/3.sql
@@ -1,3 +1,3 @@
 SELECT last_name || '-' || first_name AS full_name
 FROM actor
-WHERE first_name LIKE 'W%' OR last_name LIKE '%H';
+WHERE first_name LIKE 'W%' OR last_name LIKE '%h';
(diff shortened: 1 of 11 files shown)


qqqwww123cdv/hometask-5/3.sql
```diff
@@ -0,0 +1,3 @@
+SELECT last_name || '-' || first_name AS full_name
+FROM actor
+WHERE first_name LIKE 'W%' OR last_name LIKE '%H';
```
jekapulpit: I don't think that the anyone's last name ends on `H`, it's `h` probably 😄

qqqwww123cdv/hometask-5/4.sql
```diff
@@ -0,0 +1,3 @@
+SELECT DISTINCT rental_duration, COUNT(title) AS in_each_group
+FROM film
+GROUP BY 1;
```
jekapulpit: let's be more clear in the grouping section here, just a number brings some confusion

qqqwww123cdv/hometask-5/4.sql
```diff
@@ -0,0 +1,3 @@
+SELECT DISTINCT rental_duration, COUNT(title) AS in_each_group
```
jekapulpit: why do we need `DISTINCT` here?

qqqwww123cdv/hometask-5/5.sql
```diff
@@ -0,0 +1,5 @@
+SELECT name, COUNT(film_id) AS number_of_films
```
jekapulpit: let's specify column names in front of fields if we have more than one table in our selection

jekapulpit (changes requested):

jekapulpit (dismissed):

jekapulpit (approved):

qqqwww123cdv:
silitski_hometask/5

PR #138, opened 2021-11-22
Skripchenkoveta/hometask 5

skripchenkoVeta:


skripchenkoveta/hometask-5/4-3.sql
```diff
@@ -0,0 +1 @@
+SELECT CONCAT(last_name, '-', first_name) FROM actor WHERE first_name LIKE 'W%' OR last_name LIKE '%H'
```
jtethras: last_name like '%H' does not add any results, because all the surnames ends with h (lowercase). Please fix. The fix should be case insensitive, so no matter surname ends with h or H - this record should be added to the result

skripchenkoveta/hometask-5/4-5.sql
```diff
@@ -0,0 +1 @@
+SELECT name, COUNT(film_id) FROM film JOIN language ON language.language_id = film.language_id GROUP BY name;
```
jtethras: works as should work, but please be more specific when referring to fields, because when I read this statement and reach `GROUP BY name` section I don't understand whether it's `film.name` or `language.name`. I know `film` does not have a `name` field - it has `title` field, but I don't want to make this research, I want to read the statement and have no questions :)

`SELECT language.name, COUNT(*) FROM film JOIN language ON language.language_id = film.language_id GROUP BY language.name;`
or
`SELECT lang.name, COUNT(*) FROM film JOIN language as lang ON lang.language_id = film.language_id GROUP BY lang.name;`

skripchenkoveta/hometask-5/4-7.sql
```diff
@@ -0,0 +1,2 @@
+ UPDATE film SET rating = 'PG-13' length = 143 WHERE title = 'The Great Gatsby';
+
```
jtethras: oops what's with the line?..

skripchenkoveta/hometask-5/4-8.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM film CASCADE WHERE rating = 'PG-13';
```
jtethras: this statement causes error 
`update or delete on table "film" violates foreign key constraint "film_actor_film_id_fkey" on table "film_actor"`

jtethras (changes requested):

skripchenkoveta/hometask-5/4-7.sql
skripchenkoVeta: haha, i'm sorry
now I don't have a rubocop, so I fix it myself from memory:)

skripchenkoveta/hometask-5/4-8.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM film WHERE title = 'The Great Gatsby';
```
jtethras: Don't forget about cascade relations with other tables!
image

jtethras (changes requested):

jtethras (approved):
Nice!

PR #143, opened 2021-11-23
Romuashi/Homework-5

Romuashi:
Work with database.


Romuashi/hometask-5/4-1.sql
```diff
@@ -0,0 +1,2 @@
+SELECT * FROM actor
+WHERE last_name = 'Zellweger';
```
bedyan: add empty line at the end of each file where it is not added

Romuashi/hometask-5/4-8.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM film cascade WHERE title = 'Gorko';
```
bedyan: this will not work if I select existing film
Screenshot from 2021-11-23 21-08-31

bedyan (changes requested):

bedyan (approved):

PR #150, opened 2021-11-25
Matveibos-hometask-6

Matveibos:


Matveibos/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
+
+require_relative './my_hash'
+require 'pry'
```
vol6tmetr: Do you use it anywhere?

Matveibos/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+require 'rspec'
+
+describe MyHash do
+  subject(:my_hash) { described_class.new }
+
+  describe '#[]=' do
+    context 'when value does not exist' do
+      it 'return pair key-value' do
+        expect(my_hash[:key1] = '').to eq('')
```
vol6tmetr: It's a bad practice to setup test variables inside of a test. Check let variables and how to use them in RSpec tests

Matveibos/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+
+    context 'when value exist' do
+      before do
+        subject[:key1] = 54
+      end
+
+      it 'replace value' do
+        expect(my_hash[:key1]).to eq(54)
+        expect(my_hash[:key1] = 98).to eq(98)
+        expect(my_hash[:key1]).to eq(98)
```
vol6tmetr: https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/change-matcher
Use this matcher for the test

Matveibos/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+      end
+    end
+  end
+
+  describe '#delete' do
+    before do
+      my_hash[:key] = 34
+    end
+
+    context 'when the key exist' do
```
vol6tmetr, line 65: Why there is a context with the same name is inside of this context?

Matveibos/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+        end
+      end
+    end
+  end
+
+  describe 'clear method' do
+    it 'return []' do
+      my_hash[:a] = 82
+      my_hash.clear
+      expect(my_hash[:a]).to eq(nil)
```
vol6tmetr: I think it is better to test that after clearing you have just an empty Hash

Matveibos/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+      context 'when the wrong number of parameters was passed' do
+        it 'raises ArgumentError' do
+          expect { subject.delete }.to raise_error(ArgumentError)
+        end
+      end
+    end
+  end
+
+  describe 'clear method' do
+    it 'return []' do
```
vol6tmetr: Not the best description for that test

Matveibos/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+
+      context 'when the wrong number of parameters was passed' do
+        it 'raises ArgumentError' do
+          expect { subject.delete }.to raise_error(ArgumentError)
+        end
+      end
+    end
+  end
+
+  describe 'clear method' do
```
vol6tmetr: Describe it as you've already did for other methods

Matveibos/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+    it 'return []' do
+      my_hash[:a] = 82
+      my_hash.clear
+      expect(my_hash[:a]).to eq(nil)
+    end
+  end
+
+  describe 'length and size methods' do
```
vol6tmetr: You are testing one specific method. Reference it's name in the describe block

Matveibos/hometask-6/Gemfile
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+source 'https://rubygems.org'
+
+# gem "rails"
```
vol6tmetr, line 5: If you comment something out, anywhere in the code, leave a comment why you did that. Maybe this code is not needed at the moment but might become relevant in the future or something like that
But rails gem will not be required for this task, so I don't know why this line is here

Matveibos/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,96 @@
+# frozen_string_literal: true
+
+require_relative './my_hash'
+require 'rspec'
```
vol6tmetr, line 4: Leave a line between require statements and other code

Matveibos/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+    end
+
+    context 'when value exist' do
+      before do
+        my_hash[:key1] = 54
+      end
+
+      it 'replace value' do
+        expect(my_hash[:key1]).to eq(54)
+        expect { my_hash[:key1] = 98 }.to change { my_hash[:key1] }.to(98)
```
vol6tmetr, line 23: Read this documentation more thoroughly, you can rewrite this test to one line
https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/change-matcher

Matveibos/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    context 'when key exist' do
+      it 'return value' do
+        expect(my_hash[:key]).to eq(:value1)
+      end
+    end
+
+    context 'when key does not exist' do
+      it 'return nil' do
+        expect(my_hash[:not_exist]).to eq(nil)
```
vol6tmetr, line 48: There is a bunch of useful methods in the RSpec to check values. For example you can change this test to:
```suggestion
        expect(my_hash[:not_exist]).to be_nil
```
But you solution is okay, this is just for you to know

Matveibos/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+  describe '#delete' do
+    let(:my_hash) { MyHash.new }
+    before do
+      my_hash[:key] = 34
+    end
+
+    context 'when the key exist' do
+      it 'delete key-value' do
+        my_hash.delete(:key)
+        expect(my_hash[:key]).to eq(nil)
```
vol6tmetr, line 68: You can do that using to change test block

Matveibos/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+    end
+
+    context 'when wrong number of parameters was passed' do
+      it 'return error' do
+        expect { my_hash[] }.to raise_error(ArgumentError)
+      end
+    end
+  end
+
+  describe '#delete' do
```
vol6tmetr, line 59: Add test for deletion of not existing value

Matveibos/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+        it 'raises ArgumentError' do
+          expect { my_hash.delete }.to raise_error(ArgumentError)
+        end
+      end
+    end
+  end
+
+  describe '#clear' do
+    let(:my_hash) { MyHash.new }
+    it 'return length = 0 ' do
```
vol6tmetr, line 81: I'd rewrite this to use to change block too. Create hash with one key value pair and check that `#clear` will change it's length from 1 to 0

Matveibos/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+      my_hash[:a] = 82
+      my_hash.clear
+      expect(my_hash.length).to eq(0)
+    end
+  end
+
+  describe 'when the key exist' do
```
vol6tmetr, line 88: You're describing `#length` method, so lets make this description clearer

Matveibos/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,96 @@
+# frozen_string_literal: true
+
+require_relative './my_hash'
+require 'rspec'
+# rubocop: disable Metrics/BlockLength
+describe MyHash do
+  describe '#[]=' do
+    let!(:my_hash) { MyHash.new }
```
vol6tmetr, line 8: 1. Read about difference between `let` and `let!` methods. Why do you use `let!` here and `let` in other test suites?
2. Usually you do not need to change test variables in the test. In this case you should setup `my_hash` variable properly and then test that everything is correct, rather than modifying variable in the test
3. Spacing. Leave lines between let variable and other parts of the test suite
Everything mentioned above applies to all code, written there

Matveibos/hometask-6/scrapper.rb
```diff
@@ -0,0 +1,84 @@
...
+    within(:xpath, path) do
+      page.all(tag, class: style_class).each do |title|
+        @answer << [title.text[0...200].to_s]
+      end
+    end
+  end
+end
+
+# Write to array
+class Answer
```
vol6tmetr, line 50: I don't think that we need to create a class which will simply act as an Array. Let's turn `Finder` module into a class and modify it methods to accept array from outside, populate it with the data and return back to us

Matveibos/hometask-6/scrapper.rb
```diff
@@ -0,0 +1,84 @@
...
+        csv << row
+      end
+    end
+  end
+end
+
+infomation = Answer.new
+csv_file = FileWriter.new
+
+infomation.find_images(PATH_TO_CONTAINER_SPORT, 'sport image')
```
vol6tmetr, line 74: Let's define an array of configuration which will contain path, tag, class and comment and then just traverse through this array here, rather than write everything line by line

vol6tmetr (approved):

PR #152, opened 2021-11-25
Task 2

qqqwww123cdv:
Add RSpec for myhash.


jekapulpit (approved):
let's merge it at least after fixing hound errors

PR #156, opened 2021-11-26
Mlshvch hometask 6

mlshvch:
Hometask 6


mlshvch/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    context 'when MyHash already contains given key' do
+      it 'should replace value of the key with the new one' do
+        expect(my_hash['a1'] = 2).to eq(2)
+        expect(my_hash['a4'] = 7).to eq(7)
+      end
+    end
+
+    context 'when key is not given' do
+      it 'should raise Argument Error' do
+        expect(hash[] = 4).to raise_error ArgumentError
```
StalemateInc, line 28: Pay attention to different syntax between cases when you have to match the result of an expression against something and when you have to check for side effects of expressions.
[Github reference link, but you can also find it in RSpec docs](https://github.com/rspec/rspec-expectations#built-in-matchers)

mlshvch/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+        expect(my_hash['a'] = 1).to eq(1)
+        expect(my_hash[248] = %w[12 24]).to eq(%w[12 24])
+      end
+    end
+
+    context 'when MyHash already contains given key' do
+      it 'should replace value of the key with the new one' do
+        expect(my_hash['a1'] = 2).to eq(2)
+        expect(my_hash['a4'] = 7).to eq(7)
+      end
```
StalemateInc, line 23: A much representable way to do this will be to assign a value to the key, do it again for the same value and then check the value fetched by the mentioned key. Also, one such check is enough, those two expectations are equal.

mlshvch/hometask-4/spec/my_hash_spec.rb
StalemateInc: Additionally you can also write a spec to show that you can use `nil` as a key to show the distinction between empty key and absence of key (`nil` is still an object and can be used as a key just the same way as any other object).

mlshvch/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    context 'when new key-value pair is added' do
+      it 'should increase size by 1' do
+        my_hash[rand] = rand
+        expect(my_hash.size).to eql(11)
+      end
+    end
+
+    context 'when value of key is changed' do
+      it 'should stay the same' do
+        my_hash['a1'] = 14
```
StalemateInc, line 95:
```
Failure/Error: my_hash['a1'] = 14
     
     NoMethodError:
       undefined method `last=' for ["a1", 1]:Array
       Did you mean?  last
```

Your imlpementation of hash contains a bug and I accidentally missed that before merging. Guess now you can see why specs are important 🙂. Let's fix this as a part of this PR and cover with tests.

mlshvch/hometask-4/spec/my_hash_spec.rb
StalemateInc: `hash` here is not a local variable or memoized helper, but a method call which returns an integer. This spec does not work.

mlshvch/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+      it 'should add value to MyHash, returns assigned value and increases size' do
+        expect(my_hash['a'] = 1).to eq(1)
+        expect(my_hash[248] = %w[12 24]).to eq(%w[12 24])
+      end
+    end
+
+    context 'when MyHash already contains given key' do
+      it 'should replace value of the key with the new one' do
+        expect(my_hash['a1'] = 2).to eq(2)
+        expect(my_hash['a4'] = 7).to eq(7)
```
StalemateInc, line 22: There is no need to test for two identical scenarios. Also, such spec checks that the return value of `[]=` method matches passed value, but does not check that correct value was written into the hash.

mlshvch/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+require_relative '../my_hash'
+
+describe MyHash do
+  subject(:my_hash) { MyHash.new }
+
+  describe '#[]=' do
+    context 'when key and value are given' do
+      it 'should add value to MyHash, returns assigned value and increases size' do
+        expect(my_hash['a'] = 1).to eq(1)
+        expect(my_hash[248] = %w[12 24]).to eq(%w[12 24])
```
StalemateInc, line 15: In the description I can see three checks, but only one of them is actually covered in this `it` block.

mlshvch/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    end
+  end
+
+  describe '#clean' do
+    before { 10.times { |index| my_hash["a#{index}"] = index } }
+
+    context 'when method is called' do
+      it 'should clear MyHash' do
+        my_hash.clean
+        expect(my_hash.size).to eql(0)
```
StalemateInc, line 78: Please see the [`change` RSpec matcher](https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/change-matcher).

mlshvch/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    end
+  end
+
+  describe '#size' do
+    before { 10.times { |index| my_hash["a#{index}"] = index } }
+
+    context 'when new key-value pair is added' do
+      it 'should increase size by 1' do
+        my_hash[rand] = rand
+        expect(my_hash.size).to eql(11)
```
StalemateInc, line 89: Ditto with `change` matcher, applies to the rest of specs as well.

mlshvch/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+      end
+    end
+
+    context 'when value of key is changed' do
```
StalemateInc, line 93: What is a `value of key`? Let's write less cryptic messages, even though they could be longer. Smth like "when the same key as in inserted pair already exists".

mlshvch/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+  describe '#[]' do
+    before { 10.times { |index| my_hash["a#{index}"] = index } }
+
+    context 'when key is passed' do
+      it "should return key's value" do
+        expect(my_hash['a4']).to eq(4)
+      end
+    end
+
+    context "when key haven't been given to MyHash" do
```
StalemateInc, line 42: Ditto with the messages, in the example you are passing an "a10" string as a key, it is not obvious what to expect.

mlshvch/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+      it 'should return nil ' do
+        expect(my_hash['a10']).to eql(nil)
+      end
+    end
+
+    context 'when key is missed' do
```
StalemateInc, line 48: Ditto +
```suggestion
    context 'when key is missing' do
```

mlshvch/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    context 'when key is not given' do
+      it 'should raise Argument Error' do
+        expect(hash[] = 4).to raise_error ArgumentError
+      end
+    end
+  end
+
+  describe '#[]' do
+    before { 10.times { |index| my_hash["a#{index}"] = index } }
```
StalemateInc, line 34: It's up to you, but you don't necessarily have to fill the whole hash with values to check that actually can add a value. Instead of preparing one verifying value to check for every scenario you have, you set up a whole hash. It's a little bit redundant, as it does not impacts the outcome of the spec in any way.
TL;DR: If you need to check for something not related to dense populated hash then there is no need to populate the hash.

mlshvch/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    end
+
+    context 'when key is missed' do
+      it 'should raise ArgumentError' do
+        expect(my_hash[]).to raise_error(ArgumentError)
+      end
+    end
+  end
+
+  describe '#delete' do
```
StalemateInc, line 55: When I am describing some action taken over an object, I prefer to make an action the subject, not just object.
So, in this case 
```ruby
subject { MyHash.new }
...
describe '#delete' do
...
end
```

turns into 

```ruby 
describe '#delete' do
  subject { my_hash.delete(some_key) }

  let(:my_hash) { ...create and populate hash here... }

 context 'when something' do
  it 'does a thing' do
    expect(subject).to blah_blah
  end
 end
end
```

but since it is a very simple example, I am not sure that it's worth it.

mlshvch/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env ruby
+
+# frozen_string_literal: true
+
+require 'rspec'
+require_relative '../my_hash'
+
+describe MyHash do
+  subject(:my_hash) { MyHash.new }
```
StalemateInc, line 9: Please use [`described_class`](https://relishapp.com/rspec/rspec-core/docs/metadata/described-class) instead of referencing the name of tested class directly

StalemateInc (changes requested):
Even scraper task aside, the specs for your implementation of hash don't pass 😕

mlshvch/hometask-6/scrapper.rb
```diff
@@ -0,0 +1,80 @@
...
+require 'capybara'
+require 'capybara/dsl'
+require 'selenium-webdriver'
+
+SCRAPING_WEBSITE = 'https://www.onliner.by'
+NEWS_CSS_SELECTORS = %w[widget-1-1 widget-2-1 widget-3-1 widget-4-1 widget-5-1 widget-6-1 widget-7-1 widget-8-1
+                        widget-9-1 widget-10-1 widget-11-1 widget-12-1].freeze
+CATALOG_CSS_SELECTOR = 'div.catalog-offers__item'
+PEOPLE_NEWS_SELECTOR = 'ul.b-opinions-main-2'
+# class for parsing webpage data
```
StalemateInc, line 14: Please insert newlines between outer scope constants declaration and class body declaration.

mlshvch/hometask-6/scrapper.rb
```diff
@@ -0,0 +1,80 @@
...
+      picture = block.css('img').attr('src').text
+      text = ''
+      @result << [title, picture, text]
+    end
+  end
+
+  def write_to_csv
+    file = File.open('data.csv', 'w')
+    @result.each do |data|
+      file.write(data.join("\t").concat("\n"))
```
StalemateInc, line 77: Entries in CSV files are joined by commas, not by tabs

mlshvch/hometask-6/scrapper.rb
```diff
@@ -0,0 +1,80 @@
...
+    @html.search(CATALOG_CSS_SELECTOR).each do |block|
+      title = block.at_css('.catalog-offers__title').text.strip
+      picture = block.css('img').attr('src').text
+      text = ''
+      @result << [title, picture, text]
+    end
+  end
+
+  def write_to_csv
+    file = File.open('data.csv', 'w')
```
StalemateInc, line 75: You did neither closed the file afterwards, nor used the form which does it for you.

mlshvch/hometask-6/scrapper.rb
```diff
@@ -0,0 +1,80 @@
...
+      @result << parse_news(block)
+    end
+  end
+
+  def extract_catalog_data
+    @html.search(CATALOG_CSS_SELECTOR).each do |block|
+      title = block.at_css('.catalog-offers__title').text.strip
+      picture = block.css('img').attr('src').text
+      text = ''
+      @result << [title, picture, text]
```
StalemateInc, line 70: Why do you need to store it in a variable?
```suggestion
      @result << [title, picture, '']
```
Also (I don't really know) do catalogue items have no text?

mlshvch/hometask-6/scrapper.rb
```diff
@@ -0,0 +1,80 @@
...
+    browser = Capybara.current_session
+    driver = browser.driver.browser
+    browser.visit(SCRAPING_WEBSITE)
+    @html = Nokogiri::HTML(driver.page_source)
+  end
+
+  def parse_news(block)
+    title = block.search('.b-tile-header').text.strip
+    picture = block.at_css('img')
+    picture = !picture.nil? ? picture.attr('src') : block.search('i.b-tile-bg').attr('style').text[21...-1]
```
StalemateInc, line 53: 1. `nil` is a falsey value.
2. Do not shadow your own variables.
3. Where did 21...-1 come from? 

```suggestion
    picture_link = picture ? picture.attr('src') : block.search('i.b-tile-bg').attr('style').text[21...-1]
```

mlshvch/hometask-6/scrapper.rb
```diff
@@ -0,0 +1,80 @@
...
+    text = !block.search('.b-tile-excerpt').nil? ? block.search('.b-tile-excerpt').text : ''
```
StalemateInc, line 54: Can be simplified a lot by safe navigation usage, but be sure not to overuse.
```suggestion
    text = block.search('.b-tile-excerpt')&.text || ''
```

mlshvch/hometask-6/scrapper.rb
```diff
@@ -0,0 +1,80 @@
...
+require 'csv'
+require 'capybara'
+require 'capybara/dsl'
+require 'selenium-webdriver'
+
+SCRAPING_WEBSITE = 'https://www.onliner.by'
+NEWS_CSS_SELECTORS = %w[widget-1-1 widget-2-1 widget-3-1 widget-4-1 widget-5-1 widget-6-1 widget-7-1 widget-8-1
+                        widget-9-1 widget-10-1 widget-11-1 widget-12-1].freeze
+CATALOG_CSS_SELECTOR = 'div.catalog-offers__item'
+PEOPLE_NEWS_SELECTOR = 'ul.b-opinions-main-2'
```
StalemateInc, line 13: Is there a specific reason why it is not a part of `Scraper` class?

mlshvch/hometask-6/scrapper.rb
```diff
@@ -0,0 +1,80 @@
...
+# class for parsing webpage data
+class Scraper
```
StalemateInc, line 15: A recommendation: name your classes exactly like a file name where it is placed. Will save you a couple of gray hairs when you will start working with Rails. When I import a scra**pp**er.rb file, I expect to see a `Scrapper` class, not `Scraper`, so please make it consistent.

mlshvch/hometask-6/scrapper.rb
```diff
@@ -0,0 +1,80 @@
...
+  def extract_news_data
+    NEWS_CSS_SELECTORS.each do |selector|
+      block = @html.search("div##{selector}")
+      @result << parse_news(block)
+    end
+  end
+
+  def extract_catalog_data
+    @html.search(CATALOG_CSS_SELECTOR).each do |block|
+      title = block.at_css('.catalog-offers__title').text.strip
```
StalemateInc, line 67: That's one way to do things, but IMHO Xpath could fit a little better. We can talk about it later.

mlshvch/hometask-6/scrapper.rb
```diff
@@ -0,0 +1,80 @@
...
+    extract_news_data
+    extract_catalog_data
+    write_to_csv
+  end
+
+  private
+
+  def configure_capybara
+    Capybara.register_driver(:selenium) { |app| Capybara::Selenium::Driver.new(app, browser: :firefox) }
+    Capybara.javascript_driver = :chrome
```
StalemateInc, line 36: According to the Capybara docs, there is no such thing `:chrome` driver. Did you mean `:selenium_chrome`?

StalemateInc (commented):
1. There is barely any text in any of parsed links, even in those which lead to actual articles.
2. "Opinions" section is not present in the results at all.
3. Lots of image links are not parsed correctly.

StalemateInc (commented):
In addition to comments about specs, I recommend you to get through the [betterspec.org guidelines](https://www.betterspecs.org/) - they are pretty useful.

mlshvch/hometask-6/scrapper.rb
mlshvch: 3. This is temporary kludge)

mlshvch/hometask-6/scrapper.rb
mlshvch: Nope

mlshvch/hometask-6/scrapper.rb
mlshvch: Thanks for advice

mlshvch/hometask-4/spec/my_hash_spec.rb
mlshvch: That's very interesting because Array#last has to work. I'll try to find out the reason

StalemateInc (approved):
Merging as is because of the deadline.

PR #159, opened 2021-11-28
Skripchenkoveta hometask 6

skripchenkoVeta:


skripchenkoveta/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+  let(:my_hash) { MyHash.new }
+
+  before do
+    my_hash['li'] = 'za'
+    my_hash['my'] = 13
+    my_hash[:b] = 'name'
+    my_hash[404] = 'error'
+  end
+
+  describe 'describe #initialize' do
```
jtethras, line 17: I think here you specifically test `[]` method rather than `initialize`
In `initialize` you do
```
@size = 10
@count = 0
@myhash = Array.new(@size) { [] }
```

skripchenkoveta/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+require './myhash'
+require 'rspec'
+
+# rubocop:disable Metrics/BlockLength
+RSpec.describe MyHash do
+  let(:my_hash) { MyHash.new }
+
+  before do
```
jtethras, line 10: `[]=` method should also be covered with examples

skripchenkoveta/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      it 'return error - ArgumentError' do
+        expect { my_hash(8) }.to raise_error(ArgumentError)
+        expect { my_hash[] = 'liza' }.to raise_error(ArgumentError)
+      end
+    end
+  end
+
+  describe 'describe #clear' do
+    it 'removing all elements from the hash' do
+      expect(my_hash.clear.length).to eq 10
```
jtethras, line 61: Should not be equal 10

image

skripchenkoveta/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+  end
+
+  describe 'describe #length' do
+    it 'number of elements in the hash' do
+      expect(my_hash.length).to eq 4
+    end
+  end
+
+  describe 'negative tests' do
```
jtethras, line 38: I would rather organize specs in a way that each method `describe` block contains both positive and negative cases

skripchenkoveta/hometask-6/my_hash_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    context 'all values ​​are stored strictly by key' do
+      it 'returing the correct value' do
```
jtethras, line 40: That is not a negative scenario. Negative scenario is when you do something strange to the object, and object throws an error, and you expect the error in the spec

jtethras (changes requested):
https://www.betterspecs.org/
Use present simple in examples, not present continuous , i.e.
`it 'assigns values to keys' do`
Do not duplicate block name in block comment
i.e.  `describe '#initialize' do` instead of  `describe 'describe #initialize' do`


Recommendation:
Add `.ruby-version` file where you specify ruby version used to bundle dependencies. This way other developers can understand which ruby version to use. Some ruby version managers, i.e. RVM, when you cd to the folder where .ruby-version is located, automatically switch ruby version to the one specified in `.ruby-version`.

skripchenkoveta/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      it 'return new elements in the hash' do
+        expect(my_hash[:c] = 10).to eq 10
+        expect(my_hash[1] = -3).to eq(-3)
+      end
+    end
+    it 'return key=>value' do
+      expect(my_hash['li']).to eq 'za'
+      expect(my_hash['my']).to eq 13
+    end
+    context 'when you initialize the hash incorrectly' do
```
jtethras, line 28: this context does not belong to `describe '#[]='`

skripchenkoveta/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+
+  before do
+    my_hash['li'] = 'za'
+    my_hash['my'] = 13
+    my_hash[:b] = 'name'
+    my_hash[404] = 'error'
+  end
+
+  describe '#[]=' do
+    context 'introduction of new elements key=>value' do
```
jtethras, line 18: usually context description starts with `when...`
but here you just basically call the method `[]=` and you expect this to return the value
so `context` is redundant here, because it's just a regular scenario of using the method
so it would be
```
describe '#[]=' do
  it 'returns the value' do
    expect(my_hash[:c] = 10).to eq 10
    expect(my_hash[1] = -3).to eq(-3)
  end
```

skripchenkoveta/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      it 'return new elements in the hash' do
```
jtethras, line 19: present simple! it return**s**

skripchenkoveta/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        expect(my_hash[:c] = 10).to eq 10
+        expect(my_hash[1] = -3).to eq(-3)
+      end
+    end
+    it 'return key=>value' do
```
jtethras, line 24: here the example desc is wrong. `it` always refer to what you are describing. You should describe `[]=` method, but below you test `[]` method
If you want to test that once you assigned the value to a key you can then retrieve it, try that:
```
it 'assigns value to key' do
  expect(my_hash['new_key']).to eq null
  my_hash['new_key']  = 'some_value'
  expect(my_hash['new_key']).to eq 'some_value'
```

skripchenkoveta/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    context 'when read by key' do
+      it 'return read by key=>value' do
+        expect(my_hash[:b]).to eq 'name'
+      end
+    end
+  end
+
+  describe '#delete' do
+    it 'remove element from the hash' do
+      expect(my_hash.delete(404)).not_to eq 'error'
```
jtethras, line 45: I guess you can use `.to be_null` here

skripchenkoveta/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      expect(my_hash.delete(:b)).not_to eq 'name'
+      expect(my_hash.delete('li')).not_to eq 'za'
+    end
+    context 'when pass the wrong number of arguments in a method' do
```
jtethras, line 49: oh, in context we use present continuous! `when passing wrong number...`

skripchenkoveta/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      it 'return error - ArgumentError' do
+        expect { my_hash.delete }.to raise_error(ArgumentError)
+        expect { my_hash.delete[1] }.to raise_error(ArgumentError)
+        expect { my_hash.delete }.to raise_error(ArgumentError)
+      end
+    end
+  end
+
+  describe '#length' do
+    it 'number of elements in the hash' do
```
jtethras, line 59: it what?

jtethras (changes requested):
Please improve specs! I left some comments to `[]=` method, but there are similar fields for improvement below as well, please go through the whole spec

Also please remove old spec from /task-6

skripchenkoveta/hometask-4/spec/my_hash_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+  describe '#[]' do
+    context 'when read by key' do
+      it 'return read by key=>value' do
+        expect(my_hash[:b]).to eq 'name'
+      end
+    end
+  end
+
+  describe '#delete' do
+    it 'remove element from the hash' do
```
jtethras, line 44: Below you test that `delete` method returns null. 
If you want to test that it removes element from the hash, show that after you call the `delete` method this key-value pair does not present anymore.

PR #161, opened 2021-11-28
NastassiaBon/Hometask-6

NastassiaBon:


NastassiaBon/hometask-4/my_hash/my_hash_RSpec.rb
```diff
@@ -0,0 +1,98 @@
...
+require 'rspec'
+require 'pry'
+
+RSpec.describe MyHash do
+  subject(:my_hash) { described_class.new }
+
+  describe '#[]' do
+    before { my_hash[:key] = 2 }
+
+    context 'when key is missed' do
```
Kaddyluck, line 15: when the key is missed - how it can return the key's value?

NastassiaBon/hometask-4/my_hash/my_hash_RSpec.rb
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env ruby
+
+# frozen_string_literal: true
+
+require_relative 'my_hash'
+require 'rspec'
+require 'pry'
+
+RSpec.describe MyHash do
+  subject(:my_hash) { described_class.new }
```
Kaddyluck, line 10: it's a bad idea to have one global variable for all tests, better to use dedicated for each one

NastassiaBon/hometask-4/my_hash/my_hash_RSpec.rb
```diff
@@ -0,0 +1,98 @@
...
+      it 'should show ArgumentError' do
+        expect { my_hash[] }.to raise_error ArgumentError
+      end
+    end
+  end
+
+  describe '#[]=' do
+    context 'when key and value are set' do
+      it 'should add value to MyHash' do
+        expect(my_hash.size).to eq(1)
```
Kaddyluck, line 37: Why 1? Because you have added value in the previous test? But if I run this test separately - it will fall

NastassiaBon/hometask-4/my_hash/my_hash_RSpec.rb
```diff
@@ -0,0 +1,98 @@
...
+      end
+    end
+  end
+
+  describe '#clean' do
+    before do
+      my_hash[:a] = 1
+      my_hash[:b] = 2
+      my_hash[:c] = 3
+    end
```
Kaddyluck, line 76: This is a good example, how you need to set variables

Kaddyluck (approved):

PR #164, opened 2021-11-28
trueananas17/hometask-6

prakofyeu:


kukharava:
Small note: I asked Egor to test the Array class methods from the previous hometask, as the Hash class was not implemented correctly.

PR #189, opened 2021-12-15
VALIKKKKK/hometask-5

VALIKKKKK:


VALIKKKKK/hometask-5/4.10.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE film 
+ADD column_name TEXT
```
ShrewHools: add default value

VALIKKKKK/hometask-5/4.8.sql
```diff
@@ -0,0 +1,2 @@
+DELETE FROM film 
+WHERE title = 'It'
```
ShrewHools: Cleverly) but in these cases, you had to delete all the dependent records. _film_category_ record for example
See also **WITH** operator: https://postgrespro.ru/docs/postgresql/9.5/queries-with

ShrewHools (approved):

PR #198, opened 2021-12-23
KarinaKlevzhits/hometask6-2

KarinaKlevzhits:


KarinaKlevzhits/hometask_4/.rubocop.yml
```diff
@@ -0,0 +1,2 @@
+Metrics/BlockLength:
+   IgnoredMethods: ['describe', 'context']
```
Mat1s: In that place should be added new line

KarinaKlevzhits/hometask_4/Gemfile
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+source 'https://rubygems.org'
+
+git_source(:github) { |repo_name| "https://github.com/#{repo_name}" }
+
+# gem "rails"
```
Mat1s: This comment should be removed

KarinaKlevzhits/hometask_4/spec/lib/my_hash_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      before { my_hash.delete(:key) }
+
+      it { expect(my_hash[:key]).to eq(nil) }
+    end
+
+    context 'when key does not exists' do
+      it { expect(my_hash.delete(:key2)).to eq(nil) }
+    end
+
+    context 'when wrong parametrs in method delete' do
```
Mat1s: It's better to rename to 'when try to delete without key as argument'

KarinaKlevzhits/hometask_4/spec/lib/my_hash_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    context 'when key exists' do
+      it { expect(my_hash[:key]).to eq('hello') }
+    end
+
+    context 'when key does not exists' do
+      it { expect(my_hash[:key2]).to eq(nil) }
+    end
+  end
+
+  describe '#delete' do
```
Mat1s, line 33: Please add one more test to cover result deleting of existing key

KarinaKlevzhits/hometask_4/spec/lib/my_hash_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    context 'when key does not exists' do
+      it { expect(my_hash.delete(:key2)).to eq(nil) }
+    end
+
+    context 'when wrong parametrs in method delete' do
+      it { expect { my_hash.delete }.to raise_error(ArgumentError) }
+    end
+  end
+
+  describe '#clear' do
```
Mat1s, line 52: If I try to cover this method I'll add 3 test:
1. cover result of clear hash( expect(my_hash.clear).to eq({}) );
2. cover what my hash return after clear( expect(my_hash).to eq({}) ); 
3. cover length of hash after clear(you've done this, but rename 'it' to something like 'return length of my hash equal 0');
4. cover that key-value is unavailable after clear(that also you realized, and try to rename to 'does not return value for key 2').
Change contexts to 'when my hash cleared'(2,3,4) and 'when my hash is clearing'(1)

KarinaKlevzhits/hometask_4/spec/lib/my_hash_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+    end
+  end
+
+  describe '#delete' do
+    before { my_hash[:key] = 'ruby' }
+
+    context 'when key exists' do
+      it 'delete pair of key and value' do
+        expect { my_hash.delete(:key) }.to change(my_hash, :length).by(-1)
+        expect(my_hash[:key]).to be_nil
```
Mat1s, line 39: It's not best practice to union 2 expectations, but it's ok.

KarinaKlevzhits/hometask_4/spec/lib/my_hash_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+      end
+    end
+
+    context 'when key does not exists' do
+      it { expect(my_hash.delete(:key2)).to eq(nil) }
+    end
+
+    context 'when try to delete without key as argumen' do
```
Mat1s: forgot 't' at the end of argumen

KarinaKlevzhits/hometask_4/spec/lib/my_hash_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+      my_hash[3] = 3
+      my_hash[4] = 4
+      my_hash[5] = 5
+    end
+
+    context 'when object is valid' do
+      it { expect(my_hash.length).to eq(5) }
+    end
+
+    context 'when not valid data' do
```
Mat1s: I think in this case we need to skip context

KarinaKlevzhits/hometask_4/spec/lib/my_hash_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+  describe '#length' do
+    before do
+      my_hash[1] = 1
+      my_hash[2] = 2
+      my_hash[3] = 3
+      my_hash[4] = 4
+      my_hash[5] = 5
+    end
+
+    context 'when object is valid' do
```
Mat1s: In that situation we can skip context too, because we usually add context when initial condition is different. Condition for that 2 contexts are the same.

Mat1s (approved):

